PR #4, opened 2015-04-28
AIMS-112 (and other queue items) A first, incomplete, stab at queueing, ...

afstanton:
...for Jaron to review.


jaronkk:
## Sneakers gem
Make sure to update the sneakers gem to 1.0.4!  There was a bug with using the Maxretry handler with newer versions of RabbitMQ.  I submitted a pull request to fix the bug that got merged in and is available with the 1.0.4 version of sneakers.  Without that version, any failing jobs will retry infinitely, rather than going to the error queue after 5 failures.

## Worker
You'll need to create a worker.

See my worker at https://github.com/ndlib/honeycomb/blob/image-styles/app/workers/honeypot_image_worker.rb
It inherits the ActiveJob wrapper for sneakers, which is important to making sure it runs ActiveJob jobs correctly.

### Maxretry
I'm using the Maxretry handler from Sneakers.  This handler will use the base queue ("honeypot_images" in this example), and will create two additional queues.  <queue name>-retry and <queue name>-error.  If a job encounters an error, it will be put into the retry queue, and after 60 seconds it will be returned to the main queue.  If it fails 5 times, it will be placed in the error queue and will no longer be processed.

### worker configuration
```ruby
from_queue 'honeypot_images',
  handler: Sneakers::Handlers::Maxretry,
  workers: 1,
  threads: 1,
  timeout_job_after: 60,
  heartbeat_interval: 2,
  prefetch: 1,
  ack: true,
  arguments: {
    :'x-dead-letter-exchange' => 'honeypot_images-retry',
  },
  routing_key: ['honeypot_images']
```

maxretry config: Setting the x-dead-letter-exchange is required for the Maxretry handler to work.

workers: I don't believe the workers option has any effect at the worker level, so the number of workers can only be configured in the main sneakers.rb initializer.

from_queue 'honeypot_images' should be changed to match the name of the queue you're using.
routing_key: I'm not sure this is necessary, or what effect it has.

More configuration information here: https://github.com/jondot/sneakers/wiki/Configuration

### #work method
Modifying the work method is not necessary, but I did it to make sure Airbrake was notified during errors.  The logger.info call is not necessary

## Main configuration

config/initializers/sneakers.rb
```ruby
Sneakers.configure({
  handler: Sneakers::Handlers::Maxretry,
  amqp: Rails.application.secrets.sneakers["amqp"],
  vhost: Rails.application.secrets.sneakers["vhost"],
  workers: 1,
  heartbeat: 5,
  exchange: 'honeycomb',
  exchange_type: 'topic',
  routing_key: ['honeypot_images'],
  durable: true,
  log: 'log/sneakers.log',
})
```

`exchange: 'honeycomb'` If this isn't set, sneakers will default to an exchange named "sneakers".  Set this to whatever you want.
`routing_key: ['honeypot_images']` Once again, I'm not sure if this is necessary.

## Rake task

I wrote a set of rake tasks to run sneakers in the background: https://github.com/ndlib/honeycomb/blob/image-styles/lib/tasks/sneakers.rake

I also hardcoded my worker into the run task: https://github.com/ndlib/honeycomb/blob/image-styles/lib/tasks/sneakers.rake#L33

`rake sneakers:run` will start sneakers in the foreground.
`rake sneakers:start` will start sneakers in the background.
`rake sneakers:restart` will restart sneakers if it's running in the background (You'll need to do this whenever you make changes to your sneakers configuration, workers or jobs)

jaronkk:
It looks like ItemDataWorker and RequestsDataWorker are nearly identical.  I think you could consolidate them into a single "DefaultWorker" or something like that.  From what I understand, the only reason you need separate workers is if they're operating on different queues.  ActiveJob will make sure that each entry in the queue gets processed by the correct job.

app/jobs/get_item_data_job.rb
```diff
@@ -0,0 +1,25 @@
+class GetItemDataJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(barcode)
+Rails.logger.info "Got here"
```
jonhartzler: Is there a compelling reason to leave logger lines in?

app/jobs/get_item_data_job.rb
```diff
@@ -0,0 +1,25 @@
...
+        AddIssue.call(user_id, barcode, "Item not found.")
+        item.destroy!
+        item = nil
+      elsif data["status"] == 401 # Unauthorized - probably bad key
+        AddIssue.call(user_id, barcode, "Unauthorized - Check API Key.")
+        item = nil
+      elsif data["status"] == 599 # Timeout
+        AddIssue.call(user_id, barcode, "API Timeout.")
+        item = nil
+      end
```
jonhartzler: What is the 500 case?  

What about generic other status?

app/services/api_get_item_metadata.rb
```diff
@@ -14,19 +14,31 @@ def get_data!
...
+        {
+          "title" => raw_results["results"]["title"],
+          "author" => raw_results["results"]["author"],
+          "chron" => raw_results["results"]["description"],
+          "bib_number" => raw_results["results"]["bib_id"],
+          "isbn_issn" => raw_results["results"]["isbn_issn"],
+          "conditions" => raw_results["results"]["condition"],
+          "call_number" => raw_results["results"]["call_number"]
+        }}
+    rescue Timeout::Error => e  # Status code 599 is a MS defined code for network timeout
```
jonhartzler: Is 599 the only error we want to trap here?

app/jobs/get_item_data_job.rb
afstanton: Nope. Just forgot to take it out. I will.

app/jobs/get_item_data_job.rb
afstanton: Good point. I'll add those.

app/services/api_get_request_list.rb
```diff
@@ -18,7 +19,7 @@ def get_data!
       if !res["barcode"].blank?
         criteria_type = "barcode"
         criteria = res["barcode"]
-        item = GetItemFromBarcode.call(res["barcode"]) # Hack to make sure data is present for display.
+        item = GetItemFromBarcode.call(@user_id, res["barcode"]) # Hack to make sure data is present for display.
```
jonhartzler: Is this comment still correct?

afstanton: I'll check.

app/services/api_get_item_metadata.rb
afstanton: Timeout is a predictable potential error, so worth trapping and coding for. I don't know what other errors might happen here, so swallowing them up probably isn't desirable.

jonhartzler: Right so the idea is that a 500 at this point is an error that should be trapped with the general error and notification trapping so that the problem can be diagnosed and fixed ?

afstanton: 500 is an application error on the API side, not an error on this side. (599 is how I'm representing a timeout, it's not actually being sent by the API server.) The 500 should be handled in the job.

PR #11, opened 2015-06-24
More robust spec request stubbing

jaronkk:
Created spec helper to generate api urls based off the api configuration for the test environment.  This will prevent specs from breaking if the secrets.yml configuration is changed.


jgondron:
:+1:

PR #12, opened 2015-06-24
Improve spec performance

jaronkk:
This primarily optimizes the database cleaner to decrease the amount of time it takes to run the test suite.  In practice this cut the overall run time in half for me.


jgondron:
:+1:

PR #13, opened 2015-06-24
AIMS-209: Stock Items to Tray for invalid item throws error

jgondron:
Why: Application was throwing an exception anytime the user tried to stock an item to tray with an invalid item.
How: An FK constraint on the ActivityLog model was preventing the stubbed out item from being deleted. Since the log should keep track of all changes, deletes included, I removed the constraints and updated the application to additionally create a log entry when an Item is deleted and when it is updated using data from the API.

### DIFF
--- a/app/services/destroy_item.rb
+++ b/app/services/destroy_item.rb
@@ -1,0 +1,17 @@
+class DestroyItem
+  attr_reader :item, :user
+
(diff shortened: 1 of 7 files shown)


jaronkk:
:+1:

rfox2:
:+1: Looks great Justin

PR #14, opened 2015-06-25
AIMS-113 Reduce activity log insert to once per restock event

rfox2:
**WHAT** There was some redundant object creation in the AssociateShelfWithItem service class. It was running the association code whether or not the association already existed and this was duplicating work as well as activity logging. I added a test to avoid that duplication.

**HOW** The test checks whether the association already exists, and if so, simply restocks the item.


jaronkk:
:+1:

jgondron:
:+1:

PR #16, opened 2015-06-25
AIMS-200: Change handling of unstocked/shipped items in Items to Tray form

jgondron:
Why: Currently, there is no way to differentiate between stocked, unstocked, and shipped items, and the Unstock button is always available even if an item has been shipped or already unstocked.
How: Added a status column to the form and changed the Unstock button to only appear when an item is stocked


jaronkk:
:+1:

rfox2:
:+1:

PR #18, opened 2015-06-25
API refactoring

jaronkk:
This is some groundwork changes to the API calls that are needed before they can easily be moved to the background.  I didn't change any business logic as part of this pull request.

All API calls now return an ApiResponse which contains the status and the body of the response.  This response object can answer simple questions like whether the response was a success (200) or error.
The API handler will catch timeout errors and still return an ApiResponse.  Previously each API service was trapping their own timeout errors.
Disabled network request caching in the development and test environments.
Moved item metadata synchronization into its own service class.
Added specs for all API requests.
Updated the overall test suite order to be random, to help expose any test order dependencies.

### DIFF
--- a/app/models/api_response.rb
+++ b/app/models/api_response.rb
@@ -1,0 +1,28 @@
+class ApiResponse
+  attr_reader :status_code, :body
+
+  def initialize(status_code: status_code, body: body)
+    @status_code = status_code
+    @body = body.with_indifferent_access
+  end
+
+  def success?
+    status_code == 200
+  end
+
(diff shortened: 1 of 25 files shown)


jgondron:
Theres a bit of duplication of efforts in constructing the api urls between ApiHandler and ApiHelper. Would prefer to consolidate this.

jaronkk:
@jgondron I moved most of the API url construction into the ApiHandler class

jgondron:
:+1:

PR #21, opened 2015-06-26
Aims 212 item metadata status

jaronkk:
This is more groundwork for AIMS 212.

It adds a metadata status and timestamp which is used to determine if the item can be added to a tray.


jgondron:
:+1:

rfox2:
:+1:

PR #22, opened 2015-06-29
Fixed active requests retrieval

jaronkk:
The API refactoring introduced a bug in GetRequests.  This resolves that error.


jgondron:
:+1:

PR #23, opened 2015-06-29
AIMS-168: Remove login message

jgondron:
Why: User was getting a login required message after successfully logging in.
How: Eliminated the flash message by requiring the user to always login before trying any actions


jaronkk:
:+1:

rfox2:
:+1:

PR #24, opened 2015-06-29
Aims 212 background

jaronkk:
If an item metadata sync fails, it needs to be queued to run asynchronously in the background.  

Sync requests made in the foreground will timeout after 3 seconds.  If the foreground request fails it will queue a background request which is not under the same timeout restrictions.

I updated Rails to get ActiveJob to work as expected, since it was failing in rc1.  That caused a schema update when I ran the migrations

This will complete AIMS-212.


app/services/api_get_item_metadata.rb
```diff
@@ -1,15 +1,33 @@
...
   end
 
   def get_data!
-    ApiHandler.get(:record, barcode: barcode)
+    ApiHandler.get(action: :record, params: { barcode: barcode }, connection_opts: connection_opts)
+  end
+
+  private
+
+  def connection_opts
```
jgondron: Didn't we want to make these configurable?

jaronkk: I've moved it to a new file config/settings.yml

jgondron:
:+1:

rfox2:
:+1:

PR #25, opened 2015-06-30
AIMS-231 Show the metadata status of an item if it failed to load

jaronkk:
AIMS-231 - If an item that hasn't successfully synced with the API is stocked onto a tray, we need to display the status.

This will display the status of the metadata sync, and will highlight the row if it's not found or not for the annex.


rfox2:
:+1: Yay for i18n!

jgondron:
:+1:

PR #26, opened 2015-06-30
Aims 228 Add a scheduled task to ensure sneakers is running

jaronkk:
This adds a scheduled task (using the whenever gem) to ensure that the sneakers workers are running on the server.  It checks that the pid file is present and that the specified process id is running.


rfox2:
:+1:

jgondron:
:+1:

PR #27, opened 2015-06-30
Aims 181 remove request function

rfox2:
**What** Added functionality to allow users to remove (delete) requests for various reasons

**Why** They needed a way to arbitrarily reject requests that could not be fulfilled at the Annex and bounce them back to either Hesburgh cataloging and circ, or ILL.

**How** I created two service classes; one to delete requests, and the other to interact with the API to archive the request so that it won't show up in the active request list until the request issue is resolved. I also modified the request model so that it will also delete any associated "match" records. The button on the index.html.haml page that lists requests now has a button on it that is in a form that points to the appropriate controller method. The button is also tied to a confirmation message to avoid accidental deletions.

This is what the button looks like:

screen shot 2015-06-30 at 3 19 47 pm

### DIFF
--- a/app/controllers/requests_controller.rb
+++ b/app/controllers/requests_controller.rb
@@ -1,16 +1,13 @@
 class RequestsController < ApplicationController
-
   def remove
     request = Request.find(params[:request_id])
-    if (DestroyRequest.call(request, current_user))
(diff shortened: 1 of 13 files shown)


jgondron:
:+1:

jaronkk:
:+1:

PR #29, opened 2015-06-30
AIMS-221: Change IMS to return error if sublibrary != "Annex"

jgondron:
Why: The user needs to be notified immediately that the item should not be trayed if the sublibrary code is not marked for Annex.
How: SyncItemMetadata can now test for errors in the data even if the query is successful. When an item is not marked for annex, it will return a not_for_annex error which will trickle up to the user as an immediate "Item # not found" error.

NOTE: After merging this, you will no longer be able to pull books from the shelf for test data. You will need to find barcodes that have sublibrary codes of "ANNEX"

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -16,6 +16,7 @@ class Item < ActiveRecord::Base
     "pending",
     "complete",
     "not_found",
+    "not_for_annex",
     "error",
   ]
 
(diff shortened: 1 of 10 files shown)


jaronkk:
Address as many hound errors as are reasonable, then :+1:

rfox2:
:+1: Ditto what Jaron said

PR #30, opened 2015-06-30
AIMS-213 run the stock item api request in the background.

jaronkk:
This adds an additional worker for processing non-metadata jobs, and it moves the ApiPostStockItem service to the background


rfox2:
:+1:

jgondron:
:+1:

PR #31, opened 2015-06-30
AIMS-181 Bug fix for logging requests that don't have an item barcode

rfox2:
**WHAT** Bug fix to compensate for requests not attached to an item via barcode

**WHY** Request deletion needs to be logged even before an item match has been found

**HOW** I changed the destroy method so that it makes a clone of the request object, which is otherwise frozen, and sets the barcode to equal the trans value, which is a more reliable identifier for requests anyway.

### DIFF
--- a/app/services/destroy_request.rb
+++ b/app/services/destroy_request.rb
@@ -11,8 +11,10 @@ class DestroyRequest
   end
 
   def destroy
+    request_copy = request.clone
+    request_copy.barcode = request_copy.trans
     status = request.destroy!
-    LogActivity.call(request, "Removed", nil, Time.now, user)
+    LogActivity.call(request_copy, "Removed", nil, Time.now, user)
     status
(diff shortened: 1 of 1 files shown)


jaronkk:
:+1:

jgondron:
I'd prefer if we just add a trans field and populate the trans value there instead of mixing this in with the barcode field. This table is not normalized very well to begin with, so let's make it worse!

But seriously, we need to refactor this, and, ideally, both barcode and trans id would not be needed any longer when we do. You would only need the object types, ids, and activities. To get other information you would retrieve it from other tables (meaning we would also have to do soft deletes as part of that refactoring). So it would be better to just keep this data separate now to keep the data consistent, and we can figure out the best long term solution when we refactor.

jgondron:
:+1: as long as we refactor all of this at some point

PR #32, opened 2015-06-30
AIMS-237: Hide unstock for unstocked items in Item->Shelf

jgondron:
Why: In Item Direct to Shelf, when unstocking an item there is no feedback to the user that the item is now unstocked, and the unstock button is still available.
How: 
-Added status code to the item list
-Changed the unstock button to only be shown when the status of the item is 'stocked'


jaronkk:
:+1:

rfox2:
:+1:

PR #35, opened 2015-07-01
AIMS-215 move send/scan API requests to the background

jaronkk:
This moves the scan/send API requests to a background job.

Other changes include:
Moved match processing out of the bin controller and into a service class.
Fixed a bug where batch retrieval failed when there isn't a tray and shelf associated with an item.
Move sunspot to a different port to avoid conflicting with other local solr instances.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,12 @@
 language: ruby
 rvm:
-- 2.1.4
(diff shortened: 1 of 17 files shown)


rfox2:
:+1:

jgondron:
:+1:

PR #36, opened 2015-07-02
Requeue item sync job in background

jaronkk:
The item metadata jobs were not being requeued on failure.  In order to be requeued, an exception needs to be raised.


jgondron:
:+1:

PR #37, opened 2015-07-06
Aims 197 batch creation item status

rfox2:
**What**: Enhanced item display on batch creation page to show messages based on item status, and prevent selection of item that is "unstocked" or "shipped"

**Why**: To prevent users from selecting items that are not on the shelf, but also give them information about the item status.

**How**: Modified index.html.haml page to include conditional logic based on item status (which is driven by javascript), and modified the build_request_data service class so that item metadata includes the status information.

screen shot 2015-07-06 at 11 23 13 am

screen shot 2015-07-06 at 11 23 21 am

### DIFF
--- a/app/services/build_request_data.rb
+++ b/app/services/build_request_data.rb
@@ -33,7 +33,7 @@ class BuildRequestData
 
       if !items.blank?
         items.each do |item|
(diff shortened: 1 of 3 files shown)


jaronkk:
:+1:

PR #38, opened 2015-07-06
AIMS-202 Add description information to request display for batches

rfox2:
**What**: Added request description to batch retrieval items

**How**: Modified the HAML to include the field, and added field descriptors to output

### DIFF
--- a/app/views/batches/retrieve.html.haml
+++ b/app/views/batches/retrieve.html.haml
@@ -18,11 +18,21 @@
       %td
         %ul
           - if !@match.request.title.blank?
-            %li= @match.request.title
-          - if !@match.request.author.blank?
-            %li= @match.request.author
+            %li
+              %b= "Title: "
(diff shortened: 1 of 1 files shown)


app/views/batches/retrieve.html.haml
```diff
@@ -18,11 +18,21 @@
       %td
         %ul
           - if !@match.request.title.blank?
-            %li= @match.request.title
-          - if !@match.request.author.blank?
-            %li= @match.request.author
+            %li
+              %b= "Title: "
+              %span= @match.request.title
+          - if !@match.request.title.blank?
```
jaronkk: Should this line be checking for the author rather than the title?

rfox2: Yep! Corrected. Good catch.

jaronkk:
:+1:

PR #39, opened 2015-07-06
AIMS-241 Refactored ActivityLogs

jaronkk:
AIMS-241 Refactored ActivityLogs to store a string action and JSON data of the objects involved.

Additionally, as per AIMS-230 the ship and scan logs now record the information about the request in addition to the item.


spec/services/activity_logger_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+  let(:item) { FactoryGirl.create(:item) }
+  let(:bin) { FactoryGirl.create(:bin) }
+  let(:tray) { FactoryGirl.create(:tray) }
+  let(:shelf) { FactoryGirl.create(:shelf) }
+  let(:user) { FactoryGirl.create(:user) }
+  let(:request) { FactoryGirl.create(:request) }
+  let(:api_response) { ApiResponse.new(status_code: 200, body: { status: "OK" }) }
+
+  shared_examples "an activity log" do |message|
+    def compare_user(activity_log, user)
```
jaronkk: There's no simple way to lower the branch condition size here, so I'm not going to bother.

afstanton:
I have no horse in this race, but I feel this might make ad hoc SQL queries somewhat problematic.

spec/services/activity_logger_spec.rb
rfox2: I'm not exactly sure why this is triggering a branch condition flag. :-1:  on hound!

rfox2:
:+1:  Even though travis is unhappy.

rfox2:
@afstanton Hi Aaron! The syntax for querying against JSON will be slightly more complex (and it weds us to Postgres) but overall the reduction in complexity given what they were asking for is worth it. The activity log functionality has morphed quite a bit since you put this in place.

afstanton:
I had originally considered using polymorphism for this, which would have been a lot simpler than my original design, but ad hoc queries seemed important at the time. I agree that this is a much cleaner interface than what I had.

rfox2:
@afstanton Your original schema worked great for all of the core objects: items, shelves, trays, bins, etc. But they realized recently that given what they would need to report on for the annex usage, they also needed request transactions documented, as well as a host of other time bound stats and it had quickly outpaced what your original schema could accomodate. So, we came up with a way that would make it more flexible and a lot simpler, by just recording the activity as well as the state of each object touched by the activity. The JSON store seemed like a reasonable way to go.

afstanton:
Agreed, a JSON dump of the data is entirely reasonable, given those requirements. Getting the data back out is going to be more difficult, but I suspect you already knew that. I felt obligated to mention it, though.

jaronkk:
@afstanton This is a foray for us into the jsonb datatype in postgres: http://www.postgresql.org/docs/9.4/static/datatype-json.html

With the data stored in the jsonb datatype we can query against any of the values.  For example, to retrieve all activity logs related to an item with barcode "00000019247790":
```sql
SELECT *
FROM activity_logs
WHERE data->'item'->>'barcode' = '00000019247790';
```

Essentially we're now storing a snapshot of each object in the data attribute on the activity log.  For example the ItemStocked activity data value looks like this:

```json
{
  "item":{
    "id":101,
    "chron":"",
    "title":"Mountain City / Gregory Martin.",
    "author":"Martin, Gregory, 1971-",
    "bin_id":null,
    "status":0,
    "barcode":"00000019247790",
    "tray_id":1,
    "isbn_issn":"0865475946 (alk. paper)",
    "thickness":1,
    "bib_number":"000000001",
    "conditions":null,
    "created_at":"2015-07-06T18:30:09.364Z",
    "updated_at":"2015-07-06T18:30:10.014Z",
    "call_number":"F 849 .M68 M37 2000",
    "last_ingest":"2015-07-06",
    "initial_ingest":"2015-07-06",
    "metadata_status":"complete",
    "metadata_updated_at":"2015-07-06T18:30:09.939Z"
  },
  "tray":{
    "id":1,
    "barcode":"TRAY-AL1",
    "shelved":false,
    "shelf_id":null,
    "created_at":"2015-07-06T18:29:58.535Z",
    "updated_at":"2015-07-06T18:29:58.535Z"
  }
}
```

afstanton:
Ah - I knew that PostgreSQL supported JSON natively, but wasn't familiar with the -> notation for the queries. This makes a lot more sense.

Carry on. :)

PR #40, opened 2015-07-06
Fix a bug with removing a request (problem with nested forms), 

jaronkk:
There was a bug with removing the first request in the batch list, which was caused by nested forms.  This using the link_to method to avoid creating a form.

Also adds logging of the api remove request call for AIMS-248.


rfox2:
:+1:

PR #41, opened 2015-07-06
AIMS-233 update item issues

jaronkk:
Currently issues are being added every time a metadata API error is encountered.  The following changes are a part of this pull request:

1) Rather than store a full message, store an issue_type value (i.e. not_found, not_for_annex)
2) Items should only have issues generated for the metadata statuses of "not_found" and "not_for_annex".  Remove the other AddIssue calls
3) If an open issues already exists for a given barcode and issue_type, do not add a new issue.


rfox2:
:+1:

PR #42, opened 2015-07-06
AIMS-232 AIMS-240 Missing item messaging

jaronkk:
Improves the missing barcode message and fixes an error when trying to associate a not found item with a shelf.


rfox2:
Looks like these are bona-fide test failures from travis. When the tests are fixed, :+1:

PR #43, opened 2015-07-06
Upgrade the pg gem

rfox2:
**What**: Upgraded the pg gem

**Why**: To use the latest gem, and make sure the driver is compatible with Postgres 9.4+


jaronkk:
:+1:

jgondron:
:+1:

PR #44, opened 2015-07-07
Spec performance improvements

jaronkk:
We only need the truncation database cleaner strategy for examples that use the Capybara js driver.  The change to spec_helper `if example.metadata[:js]` cuts off several seconds.

I also added a check to the solr spec helper to monitor the status of the test instance.  That cuts off a couple seconds for the full suite as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
 - cp config/secrets.yml.travis config/secrets.yml
 - cp config/database.yml.travis config/database.yml
 - psql -c 'create database travis_ci_test;' -U postgres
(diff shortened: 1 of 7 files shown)


jgondron:
:+1:

jaronkk:
I also updated the configuration so travis is now running the entire test suite.  The only thing that wasn't working initially was three examples that make use of the Capybara javascript driver.  That was fixed by [starting xvfb](http://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-GUI-(e.g.-a-Web-browser))

rfox2:
:+1:

PR #45, opened 2015-07-07
AIMS-251: Add SkippedItem to ActivityLog

jgondron:
Why: Need to track when an item has been skipped during batch processing
How: Added a skip_item method to ActivityLogger and updated BatchesController to call it when an item is skipped.


jaronkk:
:+1:

PR #46, opened 2015-07-07
AIMS-262: Don't log UnstockedItem if an item is not associated to a tray

jgondron:
Why: Adding an item to a bin currently calls UnstockItem, which will changed the status of the item and log an UnstockedItem activity. This can happen whether the Item was associated with a tray or not. If an item is not yet associated with a tray, we don't want it to log an UnstockedItem activity.
How: Changed UnstockedItem to not log if it does not have an associated tray.

### DIFF
--- a/app/services/unstock_item.rb
+++ b/app/services/unstock_item.rb
@@ -16,7 +16,9 @@ class UnstockItem
     item.unstocked!
 
     if item.save!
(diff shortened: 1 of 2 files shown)


jaronkk:
:+1:

rfox2:
:+1:

PR #47, opened 2015-07-07
AIMS-250: Add AcceptedItem to ActivityLog

jgondron:
Why: Need to track when an item has been accepted during batch processing
How: Added an accept_item method to ActivityLogger and updated BatchesController to call it when an item is accepted.


jaronkk:
:+1:

rfox2:
:+1:

PR #48, opened 2015-07-07
AIMS-165 add a cron job for retrieving requests

jaronkk:
This removes the API call every time the batches index page is visited.
The API call is moved to a cron job that runs every 15 minutes, and can be manually triggered by clicking a new "Refresh" button.

Also updated bootstrap-sass and sass-rails to fix the glyphicons.

I also refactored the GetItemFromBarcode calls to use named parameters, which I ended up not needing, but I didn't want to make the change and not commit it.

### DIFF
--- a/app/services/get_requests.rb
+++ b/app/services/get_requests.rb
@@ -1,24 +1,82 @@
 class GetRequests
-  def self.call(user_id)
-    new(user_id).get_data!
+  def self.call
+    new.get_data!
   end
 
-  def initialize(user_id)
-    @user_id = user_id
+  def initialize
   end
 
   def get_data!
-    response = ApiGetRequestList.call(@user_id)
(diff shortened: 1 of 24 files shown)


app/services/get_requests.rb
```diff
@@ -1,24 +1,82 @@
...
   end
 
+  def create_or_update_request(attributes)
+    request = Request.find_or_initialize_by(trans: attributes["trans"])
+    request.attributes = attributes
+    request.save!
+    request
+  end
+
+  def request_attributes(request_data)
```
jaronkk: This code was moved from an existing method.  Ignoring it for now as it would require a larger refactor.

jgondron:
:+1:

rfox2:
:+1:

PR #50, opened 2015-07-08
AIMS-36 Add x of y count for items from sets

rfox2:
**What**: Added text to the right of item information on the batch retrieval interface that indicates if the current item is a part of a set that needs to be pulled in order to fulfill a request. The format is "<ordinal count> of <total number in set>".

**Why**: To let the person retrieving items know how many potential matches are left in a list of items that could fulfill a request.

**How**: Added two service classes; one that returns a boolean to indicate whether the item is a member of a set, and one that returns a hash with the ordinal designation for the current item in the match record. Also, added this to the HAML file as class methods.


app/views/batches/retrieve.html.haml
```diff
@@ -12,7 +12,11 @@
   %thead
     %tr
       %th= "Request"
-      %th= "Item"
+      %th
+        %span= "Item"
+        - if IsPartOfSet.call(@match)
+          %span= "(" + GetMatchSet.call(@match)[@match.item.id] + " of " + GetMatchSet.call(@match).count.to_s
```
jaronkk: Is there a way to do this without calling the GetMatchSet service twice?  As it is, it will have to run the query and marshal the objects twice.

I would also like to see this message consolidated into an I18n translation in en.yml

app/services/is_part_of_set.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def initialize(match)
+    @match = match
+    @request = match.request
+  end
+
+  def part_of_set?
+    Match.where(request: @request).count > 1
+  end
+end
```
jaronkk: Missing a newline here

rfox2: Added line

app/views/batches/retrieve.html.haml
rfox2: @jaronkk I pushed a change that uses a variable instead of the duplicate call. How does that look?

jaronkk:
:+1:

app/services/is_part_of_set.rb
```diff
@@ -0,0 +1,16 @@
+class IsPartOfSet
```
jgondron: This seems like a query object, but since we aren't following that pattern for this project, shouldn't this just be a part of the model? Ie Match#part_of_set?

jgondron: Also, name is too generic since this is specific to Matches. Would prefer something link IsPartOfMatchSet or similar

rfox2:
Refactored service classes into one query class

jgondron:
:+1:

app/queries/match_query.rb
```diff
@@ -0,0 +1,24 @@
+class MatchQuery
+  attr_reader :relation
+
+  def initialize(relation = Match.all, match)
+    @relation = relation
+  end
+
+  def retrieve_set(match)
```
rfox2: Not going to address this

PR #51, opened 2015-07-08
AIMS-251: Log SkippedItem

jgondron:
Why: There is more than one area where an item can be skipped. Missed this on the first PR.
How: Added a log call to the scan_bin method when the user skips the item.


rfox2:
:+1:

jaronkk:
:+1:

PR #52, opened 2015-07-08
AIMS-252: Log FilledRequest

jgondron:
Why: Need to log an activity when a request has been filled
How: Anytime a match gets processed, it will now change the status of the match to completed and additionally check to see if there are any remaining matches left for the associated request. If there are none, it will flag the request as completed and log a FilledRequest record to the ActivityLogs.

### DIFF
--- a/app/services/process_match.rb
+++ b/app/services/process_match.rb
@@ -60,7 +60,8 @@ class ProcessMatch
   def complete_match
     @match.processed = "completed"
     @match.save!
-    if @match.request.remaining_matches.count == 0
+    request = RequestQuery.new(@match.request)
+    if request.remaining_matches.count == 0
       CompleteRequest.call(request: @match.request, user: user)
     end
   end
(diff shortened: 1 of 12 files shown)


rfox2:
:+1:

app/services/process_match.rb
```diff
@@ -55,4 +56,13 @@ def scan_send
   def notify_api
     ApiScanSendJob.perform_later(match: match)
   end
+
+  def complete_match
+    @match.processed = "completed"
+    @match.save!
+    request = RequestQuery.new(@match.request)
```
jaronkk: `@match.request` can be replaced with `request`.  I would also rename your current request variable to something more like request_query

app/services/process_match.rb
```diff
@@ -55,4 +56,13 @@ def scan_send
   def notify_api
     ApiScanSendJob.perform_later(match: match)
   end
+
+  def complete_match
+    @match.processed = "completed"
```
jaronkk: `@match` should be `match`

jaronkk:
After my line notes are addressed, :+1:

PR #53, opened 2015-07-08
AIMS-258, AIMS-259 Add activity loggers for api scan/send item

rfox2:
**What**: Added loggers for API scan/send ops

**How**: Added loggers to the activity logger class, added tests, and added log_activity method to api_post_deliver_item service class

### DIFF
--- a/app/services/api_post_deliver_item.rb
+++ b/app/services/api_post_deliver_item.rb
@@ -25,9 +25,9 @@ class ApiPostDeliverItem
   def log_activity(delivery_type, params, response)
     case delivery_type
     when "scan"
-     ActivityLogger.api_scan_item(item: match.item, params: params, api_response: response)
+      ActivityLogger.api_scan_item(item: match.item, params: params, api_response: response)
     when "send"
-     ActivityLogger.api_send_item(item: match.item, params: params, api_response: response)
+      ActivityLogger.api_send_item(item: match.item, params: params, api_response: response)
     end
   end
 
--- a/app/services/api_post_deliver_item.rb
+++ b/app/services/api_post_deliver_item.rb
@@ -12,11 +12,10 @@ class ApiPostDeliverItem
 
   def post_data!
     response = ApiHandler.post(action: delivery_type, params: params)
+    log_activity(response)
     if response.success?
-      log_activity(response)
       response
     else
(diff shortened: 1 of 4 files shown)


app/services/api_post_deliver_item.rb
```diff
@@ -21,6 +22,15 @@ def post_data!
 
   private
 
+  def log_activity(delivery_type, params, response)
```
jaronkk: delivery_type and params are both available through private methods by the same name, so you don't need to pass them as arguments here

app/services/api_post_deliver_item.rb
```diff
@@ -21,6 +22,15 @@ def post_data!
...
+    case delivery_type
+    when "scan"
+      ActivityLogger.api_scan_item(item: match.item, params: params, api_response: response)
```
jaronkk: there is an item private method that you can make use of, instead of having to access match.item

jgondron:
:+1:

jaronkk:
A couple things: 

1) I would like the activity logging to happen regardless of whether it is a success or failure
2) There should be examples in  api_post_deliver_item_spec.rb verifying that ActivityLogger.api_scan_item and ActivityLogger.api_ship_item are called

rfox2:
@jaronkk I think I've covered everything you wanted. Look OK?

app/services/api_post_deliver_item.rb
rfox2: Done

app/services/api_post_deliver_item.rb
rfox2: Done

app/services/api_post_deliver_item.rb
```diff
@@ -13,14 +13,25 @@ def initialize(match:)
   def post_data!
     response = ApiHandler.post(action: delivery_type, params: params)
     if response.success?
+      log_activity(response)
       response
     else
+      log_activity(response)
```
jaronkk: Can you move this to before the if statement? Then you won't need to have it in here twice.

rfox2: Done

jaronkk:
:+1:

PR #54, opened 2015-07-09
AIMS-260 Add stock item logging

rfox2:
**What**: Added logger for api interaction when item stocked

**How**: Followed the standard pattern we've set; added a new logger method plus specs, and added a logger call when an item is stocked


app/services/api_post_stock_item.rb
```diff
@@ -14,6 +14,7 @@ def initialize(item: item)
   def post_data!
     response = ApiHandler.post(action: :stock, params: params)
     if response.success?
+      ActivityLogger.api_stock_item(item: item, params: params, api_response: response)
```
jgondron: Need a spec to ensure post_data! always logs the correct activity.

jaronkk:
I would like this to log on both success and failure.

rfox2:
I'll take care of the hound stuff

spec/services/api_post_stock_item_spec.rb
```diff
@@ -11,6 +12,8 @@
       subject { described_class.call(item: item) }
 
       it "retrieves data" do
+        expect(ApiHandler).to receive(:post).with(action: :stock, params: { :item_id => item.id, :barcode=> item.barcode, :tray_code=> tray.barcode }).and_return(response)
```
jaronkk: Can you change the hash syntax, and move the with() to the next line to address the line length concern?

jaronkk:
After hound is addressed, :+1:

jgondron:
:+1:

PR #56, opened 2015-07-09
Adding quiet_assets

jeremyf:
So the Rails logger during development is not as cluttered with
asset chatter.

More information: https://github.com/evrone/quiet_assets

$ bundle


rfox2:
:+1:

jaronkk:
:+1:

PR #57, opened 2015-07-09
Removing the test directory

jeremyf:
Given that we are using RSpec, the presence of the test directory
could be very confusing.

[skip ci]


rfox2:
:+1:

jaronkk:
:+1:

PR #60, opened 2015-07-09
Compress to module function

jeremyf:
@bcad4475c9b1e82f7d2dbbdd6a9e8cc182c50288

Freezing strings that are regularly reused

The strings were being instantiated and regularly collected via
GC. Prefer that they be instantiated once.

See for additional details http://josephyi.com/freeze/

@7258ee0699af9a6ceaf8c2466c7e9e653512546c

Prefer module functions instead of objects

The app/services/is_*.rb group of objects were instantiating an object
with each call to these "simple" service objects. This instantation had
two effects:

1) Another object that would need to be garbage collected
2) Additional steps to perform a "simple" comparison

This implementation preserves the outward facing interface of the Is*
objects, reduces memory consumption and GC, and still leaves room for
leveraging full objects if the logic gets more complicated.


jaronkk:
:+1:

PR #61, opened 2015-07-09
Updating application generator configuration

jeremyf:
Since we are using RSpec auto generate spec entries if a generator is
used.

[skip ci]


jaronkk:
Fix the hound error, then :+1:

jgondron:
:+1:

PR #64, opened 2015-07-09
Asserting that '' is not a ItemBarcode

jeremyf:
This is an interesting question, is '' an ItemBarcode?

If "" is a valid ItemBarCode then an ActiveRecord::InvalidRecord
exception is thrown on this line:

https://github.com/ndlib/annex-ims/blob/5d5e303e5fb6d4faa55157f593fc9feda78ff863/app/services/get_item_from_barcode.rb#L35

If "" is not a valid ItemBarCode then a StandardError exception is
thrown.

https://github.com/ndlib/annex-ims/blob/5d5e303e5fb6d4faa55157f593fc9feda78ff863/app/services/get_item_from_barcode.rb#L21

It is important to decide what is the correct answer as we the
handling of [AIMS-271](https://jira.library.nd.edu/browse/AIMS-271)
will vary

### DIFF
--- a/app/services/is_item_barcode.rb
+++ b/app/services/is_item_barcode.rb
@@ -1,5 +1,9 @@
 module IsItemBarcode
   def self.call(barcode)
-    !(IsTrayBarcode.call(barcode) || IsShelfBarcode.call(barcode) || IsBinBarcode.call(barcode))
+    return false unless barcode.present?
+    return false if IsTrayBarcode.call(barcode)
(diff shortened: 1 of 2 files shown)


jaronkk:
"" should not be considered an item barcode

jaronkk:
:+1:

jeremyf:
@jaronkk I was afraid that would be the answer; The controller solution is now very ugly.

PR #65, opened 2015-07-09
AIMS-255: Log BatchedRequest

jgondron:
Why: Need to track anytime a request is batched
How: Modified BuildBatch to log an activity for each unique request associated with the batch (since a batch can have multiple items for one request). To do this I had to modify the association between Batch and Request to only include unique values since it's through a 1->many relationship with Match


jaronkk:
:+1:

rfox2:
:+1:

PR #67, opened 2015-07-09
AIMS-271: Handle an exception due to data validation

jeremyf:
This is not an ideal solution, but instead addresses the issue
where not providing a Barcode for the /trays/items page throws an
exception.

Closes AIMS-271

### DIFF
--- a/app/controllers/trays_controller.rb
+++ b/app/controllers/trays_controller.rb
@@ -148,15 +148,21 @@ class TraysController < ApplicationController
       thickness = params[:thickness]
     else
       flash[:error] = 'select a valid thickness'
-      redirect_to show_tray_item_path(:id => @tray.id, :barcode => barcode)
+      redirect_to show_tray_item_path(id: @tray.id, barcode: barcode, thickness: params[:thickness])
       return
     end
 
-    item = GetItemFromBarcode.call(barcode: barcode, user_id: current_user.id)
+    begin
+      item = GetItemFromBarcode.call(barcode: barcode, user_id: current_user.id)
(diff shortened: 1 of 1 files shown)


app/controllers/trays_controller.rb
```diff
@@ -148,15 +148,21 @@ def associate_item
...
       flash[:error] = 'select a valid thickness'
-      redirect_to show_tray_item_path(:id => @tray.id, :barcode => barcode)
+      redirect_to show_tray_item_path(id: @tray.id, barcode: barcode, thickness: params[:thickness])
       return
     end
 
-    item = GetItemFromBarcode.call(barcode: barcode, user_id: current_user.id)
+    begin
+      item = GetItemFromBarcode.call(barcode: barcode, user_id: current_user.id)
+    rescue StandardError
```
rbalekai: NotifyError.call(exception: e) ?

jeremyf: Good catch. I don't think we need to report this exception as it is being pushed back to the user and is something they can handle.

jaronkk: Except there are more things than what we expect that could go wrong in the GetItemFromBarcode class, right?

GetItemFromBarcode should probably raise a specific error that we can rescue and use for messaging.  Then we can use rescue StandardError to to the error notification.

jeremyf: I've run out of time, but I believe you can build from this solution.

jgondron: Agree that it should throw a specific error that's rescued for messaging

PR #69, opened 2015-07-09
Verifying that Match#bin_id is present

jeremyf:
Is this a requirement?


jaronkk:
Nope, matches (joins between requests and items) get set up when a user creates a batch, before they go to retrieve items.

The bin association happens when the user retrieves an item from a shelf, and put it in their bin to take back to where they do request processing.

PR #70, opened 2015-07-09
AIMS-270: UnstockedItem message when fullfilling a request

jgondron:
Why: For item scans, we are getting an extra UnstockedItem message in the log when the request is completed.
How: ScanItem was calling an extra UnstockItem. At this point it should be unstocked, so I removed the call. Also changed UnstockItem to not perform the update to the object or log the call if the item was already unstocked.


jaronkk:
:+1:

PR #71, opened 2015-07-09
Adding shoulda matcher to convey intentionality

jeremyf:
I had a question concerning whether or not a validation was missing. Since there is not a spec asserting whether validation is required or not, I didn't know.

Thoughts?


jgondron:
:+1:

jaronkk:
:+1:

PR #72, opened 2015-07-09
AIMS-256: Log MatchedItem

jgondron:
Why: Need to track when an item is matched
How: Added a match_item log method and changed BuildBatch to log a MatchedItem for each item added to matches.


jaronkk:
:+1:

PR #76, opened 2015-07-09
AIMS-107, AIMS-108, AIMS-109, AIMS-110 Add item detail page

rfox2:
**What**: Added a first pass at an item detail page that will show current item information including status, the history of what has happened with the object at the Annex, and how often an item was requested and shipped/scanned.

**How**

* Added a new controller method and route
* Added a new query class for activity logs with item_history and item_usage methods
* Created the item detail page
* Added some i18n for activity messages
* Created appropriate unit tests

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -26,6 +26,12 @@ class ItemsController < ApplicationController
     @item = Item.find(params[:id])
   end
 
+  def item_detail
+    @item = Item.where(barcode: params[:barcode]).take
+    @history = ActivityLogQuery.item_history(@item)
+    @usage = ActivityLogQuery.item_usage(@item)
+  end
+
   def restock
     item_id = params[:id]
     barcode = params[:barcode]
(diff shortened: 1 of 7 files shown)


rfox2:
Will address hound issues tomorrow

jgondron:
:+1:

jaronkk:
I'd like to start under the assumption that we'll use a gin index on data.  To take advantage of that we'll need to use different operators.  For now I'd like to change the where data->item->>barcode to the following: `where("data @> ?", { item: { barcode: item.barcode } }.to_json)`.  Here's a description of the jsonb operators: http://www.postgresql.org/docs/9.4/static/functions-json.html#FUNCTIONS-JSONB-OP-TABLE

I would also like to avoid repetition where possible, so doing something like:

```ruby
module ActivityLogQuery
  module_function

  def for_item(item)
    relation.where("data @> ?", { item: { barcode: item.barcode } }.to_json)
  end

  def item_history(item)
    for_item(item).
      where(action: ['StockedItem','UnstockedItem','CreatedItem','DestroyedItem']).
      order(action_timestamp: :desc)
  end
```

jaronkk:
:+1:

PR #77, opened 2015-07-10
AIMS-255: Log BatchedRequest

jgondron:
Why: Was causing duplicate BatchedRequest records for Requests that were already added to the batch.
How: Added an additional filter to limit it to just the Request ids that were added as part of the current BuildBatch#build! action.


jaronkk:
:+1:

PR #80, opened 2015-07-10
AIMS-234: Autofocus on Search within Items with Issues

jgondron:
Why: A new barcode search field was added, but we already had a search field that allowed searching by barcode. 
How: Removed the added field and gave focus to the existing Search input. Also added seed data for the issues table


jaronkk:
:+1:

PR #81, opened 2015-07-10
AIMS-273 Log ResolvedIssue.

jaronkk:
Log ResolvedIssue.  Also updated AddIssue to log the issue after it's saved (currently it's logging the issue record before it's created)


jgondron:
:+1:

PR #82, opened 2015-07-10
Log the user when adding an issue

jaronkk:
The user was not getting logged when an issue was added.  This can help in identifying who scanned the barcode that has the issue.


rfox2:
:+1:

jgondron:
:+1:

PR #83, opened 2015-07-10
AIMS-246: User access

jgondron:
Why: Need to restrict access to the application to a specified list of users
How: Added an admin field to the users table and modified the ApplicationController to check this. Essentially only administrators of the application will have any access, and they will all have access to a form to edit the users. In order to give permission to the first user you can configure an admin_user_name in the config.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,7 +13,7 @@ class ApplicationController < ActionController::Base
   end
 
   def redirect_to_unauthorized
-    raise_404("User not an admin")
+    raise_404
   end
 
   def raise_404(message = "Not Found")
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,12 +1,22 @@
 class ApplicationController < ActionController::Base
-  before_action :redirect_to_sign_in, unless: :user_signed_in?
-  before_action :redirect_to_unauthorized, unless: :user_admin?
+  before_action :check_authentication
 
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
   protected
+
+  def check_authentication
+    unless user_signed_in?
+      redirect_to_sign_in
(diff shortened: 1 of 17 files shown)


jaronkk:
Can you add a spec asserting that you receive a 404 if you're logged in as a user with admin set to false?

app/controllers/application_controller.rb
```diff
@@ -10,4 +11,20 @@ class ApplicationController < ActionController::Base
   def redirect_to_sign_in
     redirect_to new_user_session_path
   end
+
+  def redirect_to_unauthorized
+    raise_404("User not an admin")
+  end
+
```
rfox2: Should this be a 401?

config/environments/development.rb
```diff
@@ -41,6 +41,9 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+
+  # Configures a user as an administrator
+  config.admin_user_name = "jgondron"
```
rfox2: Good - none of the rest of us want this massive responsibility - you put this in pprd and prod too right?

app/controllers/application_controller.rb
jgondron: I'm following DEC, we throw 404

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,17 @@
+class UsersController < ApplicationController
+  def index
+    @users = User.all
+  end
+
+  def update
+    user_id = params["user_id"]
```
rfox2: Could we add a basic feature test for this?

config/environments/development.rb
jgondron: I can!

jaronkk:
I added an example asserting that it redirects to the sign in path if you're not logged in.

jaronkk:
:+1:

rfox2:
:+1: Let's put "mvanneve" as the admin for prod and pprd, and @jgondron as the admin in staging

PR #87, opened 2015-07-13
AIMS-246: User access

jgondron:
Why: If a user already exists, throws an unhandled exception
How: Changed to handle when a user already exists and flash an error

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -5,8 +5,12 @@ class UsersController < ApplicationController
 
   def create
     user_name = params["user_name"]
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -5,8 +5,12 @@ def index
 
   def create
     user_name = params["user_name"]
-    user = User.new(username: user_name)
-    user.save!
+    if User.exists?(username: user_name)
+      flash[:error] = "User already exists"
```
rfox2: How about a small feature test for this conditional?

rfox2:
:+1:

PR #88, opened 2015-07-14
Remove unstock button temporarily

jgondron:
Why: We want to remove this button to prevent accidents
How: Temporarily commented this out in the code.


rfox2:
:+1:

PR #90, opened 2015-07-16
Aims 284 shelf detail page

rfox2:
**What**: Added a shelf detail page, cleaned up the date display, and added a activity log query method for shelves

**Why**: They needed it for QC at the annex

**How**: There is now a new route for the reports index, where we could position a staging ground for more reporting in the IMS; The barcode field on that page now works with item and shelf barcodes automatically; Need to add a tray detail page; Touched several views to fix date formatting;


config/routes.rb
```diff
@@ -19,6 +19,7 @@
   get "trays/items", to: "trays#items", as: "trays_items"
   post "trays/items", to: "trays#scan_item", as: "scan_tray_item"
   get "trays/items/:id", to: "trays#show_item", as: "show_tray_item"
+  get "trays/detail/:barcode", to: "trays#tray_detail", as: "tray_detail"
```
jgondron: There is no tray_detail in TraysController. Should probably have something stubbed out to at least say that the report isn't available for trays.

jgondron:
:+1: after build, hound, and comments are addressed :smiley:

PR #92, opened 2015-07-21
AIMS-288 Fix sneakers restart while jobs are in the queue

jaronkk:
Removing Typhoeus in favor of Net HTTP since Typhoeus was causing issues with process forking.

Move sneakers rake methods into helper class to avoid global method definitions
Only include the api response attributes in api errors to allow some grouping of error messages for each item
Disable any request caching to prevent receiving stale data from the API server
Use TERM instead of INT to kill running sneakers processes (INT could detach processes rather than stop them depending on the configuration)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -41,8 +41,6 @@ group :application do
   # For consuming the API for items
   gem "faraday"
   gem "faraday_middleware"
-  gem "excon"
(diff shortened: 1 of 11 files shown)


rfox2:
:+1:

jgondron:
:+1:

PR #93, opened 2015-07-22
Improve quality of error information sent via Airbrake

jaronkk:
This includes a number of changes to notifications sent through Airbrake.

1.Calls using our NotifyError service don't have access to the controller params and environment.  Using the notify_airbrake method provided by the Airbrake gem allows us to send more information.
2. I updated the rake tasks to make use of Airbrake's internal handler for errors in rake tasks
3. I enhanced the NotifyError service to send more information (similar to how Airbrake handle's rake tasks)
4. Improved a handful of related specs
5. Updated the staging environment to match prod/pprd configurations so it will send error messages.


jgondron:
:+1:

rfox2:
:+1:

PR #94, opened 2015-07-27
AIMS-295 Add message body for 422 return codes

rfox2:
What: Added return message body for 422 return codes
Why: So that we can see the error message returned from Aleph in the activity log
How: Modified the external rest class so that it includes message bodies for that return code

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -464,5 +464,5 @@ DEPENDENCIES
   webmock
   whenever
 
-  BUNDLED WITH
-     1.10.5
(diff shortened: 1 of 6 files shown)


jaronkk:
Can you add an example to the ExternalRestConnection spec that verifies it returns a body for the 422 error.  Similar to the `context "500 error" do` example.

rfox2:
NO

jaronkk:
:+1:

jgondron:
:+1:

PR #96, opened 2015-07-29
AIMS-295:  Add 422 issue

jgondron:
Why: Whenever we get a 422 back from the api when stocking an item, it means that aleph encountered an error with processing/updating the item. This needs to bubble up to the user as an issue so that someone can manually address the problem in Aleph.
How: Created a new aleph_error type in issues and started creating an issue when we receive a 422 during ApiPostStockItem. The api will now respond with the error message that it received from aleph, so added this to issues table/form. We also decided to drop the fk constraint and user_id validation on issues. Issues previously required an associated user, but background operations do not have a user.


rfox2:
:+1:

PR #97, opened 2015-07-29
AIMS-298 Minor cleanup

rfox2:
What: Just a minor code revision
Why: Addressing an ABC length issue for testing purposes
What: Shortened method and removed conditionals and several branch calls


jgondron:
:+1:

PR #99, opened 2015-07-30
AIMS-297 Add more info to item detail page

rfox2:
What: Aaron requested that we indicate when an item was dissociated from a tray on the item detail page, which I did and also added several other transitions to the list.
How:  Added several actions to the item_history method and cleaned up the activity_log_query.rb file a bit.

### DIFF
--- a/app/services/activity_logger.rb
+++ b/app/services/activity_logger.rb
@@ -159,7 +159,7 @@ class ActivityLogger
       data: data,
       username: username,
       user_id: user_id,
-      action_timestamp: Time.now,
+      action_timestamp: Time.zone.now,
     )
   end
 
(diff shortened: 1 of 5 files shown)


jaronkk:
:+1:

app/services/activity_logger.rb
```diff
@@ -159,7 +159,7 @@ def log!
       data: data,
       username: username,
       user_id: user_id,
-      action_timestamp: Time.now,
+      action_timestamp: Time.zone.now,
```
jgondron: Will rails convert this to UTC when writing to the db?

jaronkk: I think Rails should have already been converting this to UTC in the database, so I don't know if this change makes any difference.  @rfox2 ?

rfox2: You're right. I'll remove it.

jgondron:
:+1:

PR #101, opened 2015-07-30
AIMS-283 Modify item status when related issue

rfox2:
What: A request came in to change the status message for items that were also on the issue list. Instead of changing the status, I opted to create a decorator that will modify the status based on whether the item is on the issue list.
How: Cleaned up item model, added decorator and specs, added query class and specs

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -28,7 +28,7 @@ class ItemsController < ApplicationController
   end
 
   def item_detail
-    @item = Item.where(barcode: params[:barcode]).take
+    @item = Item.where(barcode: params[:barcode]).take.extend(ItemWithIssue)
     if @item
       @history = ActivityLogQuery.item_history(@item)
       @usage = ActivityLogQuery.item_usage(@item)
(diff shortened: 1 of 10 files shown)


jaronkk:
I think the use of extend here could cause confusion in the future.  I'd rather see the decorator be its own class that accepts item as an argument during initialize, and then delegates everything except for status to the item. (Similar to how we use Draper decorators)

rfox2:
I'll have to look and see how that was set up. Sounds pretty clean cut though.

jaronkk:
I don't think we need to add draper since this is our only decorator right now.  Something relatively simple like this:

```ruby
class ItemWithIssue
  attr_reader :item
  
  delegate :barcode, :title, :author, to: :item
  
  def initialize(item)
    @item = item
  end
  
  def status
    if IssuesForItemQuery.call(barcode: barcode).count >= 1
      item.status + " (on issue list)"
    else
      item.status
    end
  end
end
```

You'd have to expand the delegate call to add all of the methods needed in the view.

rfox2:
Right, but I was considering just putting in a generic decorator that's 95% delegated, and targeted for the views. I think my intention was actually a presenter, now that I've written that. That's what's needed here as opposed to a decorator.

jaronkk:
:+1:

jgondron:
:+1:

PR #102, opened 2015-07-30
AIMS-264, AIMS-265 Item metadata resync

rfox2:
What: Added some code to resync the item metadata when a matching request is created, and when the item is restocked (or stocked as the case may be).
How: cleaned up several classes, modified SyncItemMetadata so that it no longer requires a user, placed calls to that routine in the relevant classes, and added several specs.


jaronkk:
### GetRequests spec
The spec needs to verify that update_item_metadata is called from create_or_update_request.

It also needs a spec verifying that the update update_item_metadata method works.  There should be an example for the case where the item already exists in the database, and where the item doesn't yet exist in the database.


### GetRequests#update_item_metadata
This method should make use of the GetItemFromBarcode service.  `Item.where(barcode: barcode)` will return an activerecord relation, so my guess is this method doesn't work as expected currently.

### SyncItemMetadata#sync_required?

The `sync_required?` logic needs to be updated.  Right now it won't check for new metadata if it has the metadata_status "complete".  It needs to be updated so that for items with the "complete" status, it will sync the metadata if it hasn't been updated within the last 24 hours.  You could update the `sync_required_based_on_status?` method to do that.  Something like:
```ruby
def sync_required_based_on_status?
  item.metadata_status != "complete" || item.metadata_updated_at < 24.hours.ago
end
```

jaronkk:
Actually, if you make the logic change to `sync_required_based_on_status?`, and with the way GetItemFromBarcode works, you wouldn't need to explicitly call SyncItemMetadata, as long as you make a call to GetItemFromBarcode.

GetItemFromBarcode will always make a call to SyncItemMetadata, so you wouldn't need to make your own call.

GetItemFromBarcode also needs to be updated to allow a nil user_id (by making the same change to the `user` method).

rfox2:
I've chosen not to use the GetItemFromBarcode service because of the way that class is implemented and instead opted to add ".take" to the active record call which avoids the problem of returning the AR relation (which I keep forgetting it does!). It'll either return nil or an item, which accomplishes the same thing. The GetItemFromBarcode class repeats a call to SyncItemMetadata as a foreground process, and we want this call to be backgrounded. For now, I think this way makes sense (but I'm always open to a counter argument!).

jaronkk:
So this is poor argument naming on my part, but the "background" argument in SyncItemMetadata doesn't queue the Sync in the background, it's actually telling SyncItemMetadata whether it's running in the background context or not.  To queue it in the background you'll want to run `  SyncItemMetadataJob.perform_later(item: item, user_id: nil)`

app/services/sync_item_metadata.rb
```diff
@@ -44,7 +44,15 @@ def sync_required?
...
-    item.metadata_status != "complete"
+    item.metadata_status != "complete" || recent_metadata_update?
+  end
+
+  def recent_metadata_update?
+    if !item.metadata_updated_at.blank?
+      item.metadata_updated_at < 24.hours.ago
+    else
+      false
+    end
```
jaronkk: Can we rename this method to something like `metadata_stale?` ?  And have it return true if metadata_updated_at is blank.

rfox2: Isn't this covered by line 47? The only time it should be blank is when the status isn't "complete" right?

jaronkk: Right, there are no logic holes in the class overall.  I'm more looking to make sure that this specific method name and logic are consistent with each other, to avoid confusing in the future if someone else makes use of this particular method.

rfox2: I've pushed up some changes. Take a look and see if it looks OK now.

jaronkk:
:+1:

jgondron:
:+1:

PR #103, opened 2015-07-31
Add a maintenance page during deploy

jaronkk:
This adds a maintenance page to public/system/maintenance.html during deployments.


jgondron:
:+1:

rfox2:
:+1: Once Mark S. has it enabled on the nginx server

PR #104, opened 2015-08-17
AIMS-303: Activity indexing

jgondron:
Why: Item/shelf detail report is extremely slow.
How: Analyzed the query that this report uses and found that adding an index on data->item->barcode and data->shelf->barcode will speed up the query substantially. Added these indices and changed the query to use the ? operator to make use of the indices.


rfox2:
:+1:

jaronkk:
:+1:

PR #106, opened 2015-08-27
AIMS-317: Request date not showing correct date

jgondron:
Why: The requested date does not reflect when the request was actually created.
How: Changed the requested field to use the request_date_time received from the api instead of todays date.


jaronkk:
:+1:

rfox2:
:+1:

PR #107, opened 2015-08-27
AIMS-310: Autofocus on search criteria to make scanner easier to use

jgondron:
Why: To simplify usage with the scanner
How: Added autofocus to the criteria box on item search.


jaronkk:
Does the scanner browser support the autofocus attribute?

jgondron:
Hard to say until I can test on staging. It seems to work within Batch retrieval (which explicitly uses autofocus), but that still could just be due to the order of the fields. The ticket won't be closed until we can confirm it works in QA

jaronkk:
:+1:

jgondron:
Pushed to staging and tested the scanner. Seems to work fine.

rfox2:
:+1: Yes, the scanner browser supports autofocus

PR #108, opened 2015-08-28
AIMS-315: Directed put away info not visible

jgondron:
Why: When scanning an item in directed put away, the info appears at the bottom. The screen on the scanner is fairly small, and this info gets cut off. Reworked the layout of this form to make all needed info visible.
How: Had to remove a label, move the info to the top, and make the paragraphs a little more compact in order to accommodate for the info, inputs, and error message.


jaronkk:
:+1:

rfox2:
:+1:

PR #109, opened 2015-08-31
AIMS-299, AIMS-281 improve items with issues table

jaronkk:
AIMS-299: Adjustments to the display issues table to limit text wrapping:

1. Removed the "Message" column, and if the message is present, display it under the type
2. Limit the width of the barcode column

AIMS-281: Fixed created date sorting
1. Added a hidden timestamp column to use as sorting data for the Created column
2. Barcode sorting appears to be working just fine.

### DIFF
--- a/app/views/items/issues.html.haml
+++ b/app/views/items/issues.html.haml
@@ -7,16 +7,15 @@
       %tr
         %th= 'Barcode'
         %th= 'Type'
-        %th= 'Message'
         %th= 'Created'
         %th= 'User'
         %th= ''
+        %th= 'Created Timestamp'
     %tbody
       - @issues.each do |issue|
         %tr
           %td= link_to issue.barcode, item_detail_path(issue.barcode)
           %td= t "issues.issue_type.#{issue.issue_type}"
-          %td= issue.message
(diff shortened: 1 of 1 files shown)


app/views/items/issues.html.haml
```diff
@@ -7,16 +7,15 @@
...
         %th= 'Created'
         %th= 'User'
         %th= ''
+        %th= 'Created Timestamp'
     %tbody
       - @issues.each do |issue|
         %tr
           %td= link_to issue.barcode, item_detail_path(issue.barcode)
           %td= t "issues.issue_type.#{issue.issue_type}"
-          %td= issue.message
```
jgondron: Are you sure this isn't used? IIRC we added this to show the aleph errors associated with item updates.

jgondron: Yeah, this was broken out into type/message, where message would have any additional details about the error received from aleph. This column is not used very often though. Perhaps we can just concat the type and message into one field? Not sure if this would negatively affect the users' ability to sort on that field or not.

jaronkk:
The message is now displayed as part of the type column: annex_ims

jgondron:
:+1:

rfox2:
:+1:

PR #110, opened 2015-08-31
AIMS-318: Too many items for a request

jgondron:
Why: If a request has many items that match, it does not show all items that it should.
How: Changed to retrieve 5,000 items per page instead of 50. There is still no way to go to next page of items (without a redesign), so if there are more than 5k items (more than one page) it will send an error message in with the hash of item data. The js that processes this will now print the error to the user to notify them that the request is likely bad.

Example:
screen shot 2015-09-01 at 3 48 27 pm

### DIFF
--- a/app/views/batches/index.html.haml
+++ b/app/views/batches/index.html.haml
@@ -29,7 +29,7 @@
         %th= 'Remove Request'
     %tbody
       - @data.each do |row|
-        %tr{"data-params" => row['item_data'].to_json}
+        %tr{"data-items" => row['item_data'].to_json, "data-error" => row['error'] }
           %td
             %div{class: "details-control btn btn-primary", "id" => 'request_' + row['id'].to_s}= 'Results'
           %td= row['rapid']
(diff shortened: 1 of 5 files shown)


app/views/batches/index.html.haml
```diff
@@ -141,6 +139,10 @@
       str += "<th>Author</th>";
       str += "<th>Chron</th>";
       $.each(json, function( index, item ) {
+        if (item['error']) {
+          str += "<div style='color: red;'><b>" + item['error'] + "</b></div>";
```
jaronkk: I think this needs to be within a <tr> and <td> tag so we're returning valid html.  It could be something like
```javascript
str += "<tr><td colspan=\"6\"><b style='color: red;'>" + item['error'] + "</b></td></tr>";
``` 

Also, in the javascript above this the <th> tags need to be contained within a <tr> as well.

app/views/batches/index.html.haml
```diff
@@ -128,29 +127,43 @@
...
+      str += "</tr>";
       $.each(json, function( index, item ) {
-        str += "<tr>";
-        if (item['status'] == 'stocked') {
-          str += "<td><input class='item' type='checkbox' name='batch[]' id='"+item['id']+"' value='"+item['id']+"' /></td>";
-        }
-        if (item['status'] == 'unstocked') {
-          str += "<td><span style='color: red'>Unstocked</span></td>";
+        if (item['error']) {
+          msg += "<div style='color: red;'><b>" + item['error'] + "</b></div>";
```
rfox2: Inline styles...hmmm. Perhaps we should have an "error color" to be consistent across error messages?

app/views/batches/index.html.haml
```diff
@@ -128,29 +127,43 @@
     } );
   };
   function format(json) {
-    console.log("my object: %o", json)
+    msg = "";
     if (json.length < 1) {
       str = "No results.";
     }
     else {
       str = "<table class='table table-striped condensed datatable'>";
```
rfox2: This is just a question: does the class "datatable" automatically trigger the data table code or do we need to invoke the Datatable() method to instantiate it? I suspect the latter. Is this where we would inject an embedded datatable that could include the filter tool?

spec/services/build_request_data_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    {
+      status: "status",
+      shelf: shelf,
+      tray: tray,
+      title: "title",
+      author: "author",
+      chron: "chron"
+    }
+  end
+  let(:requests) { [instance_double(Request, id: 1, criteria_type: "bib_number", criteria: "bib_number", **request_data)] }
```
rfox2: The ** syntax is cool! I didn't know we could do that.

spec/services/build_request_data_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:item) { instance_double(Item, id: 1, **item_data) }
+  let(:results) { instance_double(Sunspot::Search::PaginatedCollection, total_pages: 1) }
+  let(:search) { instance_double(Sunspot::Search::StandardSearch, results: results) }
+  let(:shelf) { instance_double(Shelf, barcode: "shelf_barcode") }
+  let(:tray) { instance_double(Tray, barcode: "tray_barcode") }
+
+  context "found items" do
+    before(:each) do
+      allow(SearchItems).to receive(:call).and_return(search)
+      allow(results).to receive(:each).and_yield(item)
```
rfox2: Does the each loop actually return an item object?

```ruby
request_data["item_data"] << temp
```

spec/services/build_request_data_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+
+    it "returns request data" do
+      result = described_class.call(requests)
+      expect(result[0].symbolize_keys).to include(request_data)
+    end
+
+    it "returns no item data for the request" do
+      result = described_class.call(requests)
+      expect(result[0]["item_data"]).to eq([])
```
rfox2: Would this be an empty array, or nil if there are no items returned?

rfox2:
Other potential unit tests:
* case where the submitted request param is empty or nil
* what gets returned from SearchItems.call when there are no items found? I think you may have tested this - just checking.

spec/services/build_request_data_spec.rb
jgondron: It would be [] since request_data["item_data"] is initialized as [], and only modified if items is not blank

app/views/batches/index.html.haml
jgondron: Worst. Suggestion. Evar. 

Seriously, though, meant to go back over this. Good catch.

app/views/batches/index.html.haml
jaronkk: The DataTable() function would need to be called to turn this subtable into its own datatable.  We'll worry about that when we address the filtering ticket, but yeah it will be somewhere around here.

rfox2:
If we're OK with the unit tests as they stand, then :+1:

PR #111, opened 2015-09-01
AIMS-300 save the state of the issues table

jaronkk:
Datatables has built-in support for saving the state of the table via local storage.  This saves it for a short time (5 minutes), so if the user leaves the page and then comes back, the table will be in the state that it was in when they left the page.


rfox2:
:+1:

jgondron:
:+1:

PR #112, opened 2015-09-01
AIMS-319 Track pickup location, patron institution, department, and status

jaronkk:
What: They want to be able to do reporting on requests using the patron status (Grad/Undergrad), patron institution, patron department, and pickup location for requests.

I added four new fields to the request table and set their values from the data returned by the API.

I also need the `AIMS-319-request-patron-info` branch reviewed on the API repository.


jgondron:
:+1:

rfox2:
:+1:

PR #113, opened 2015-09-02
AIMS-272: Log activity when removing a match

jgondron:
Why: If a user removes a match on the "View Current Batch" screen, it is not logging the activity and we need it for reporting
How: Added a RemovedMatch activity log and modified batches controller to log this activity any time the user removes a matched item from a batch.


rfox2:
:+1:

jaronkk:
:+1:

PR #114, opened 2015-09-02
AIMS-322 error handling and debugging for requests sync

jaronkk:
Currently when requests are loaded from the API, an error on an individual request will break the entire process and will not provide any helpful debugging information.

This adds error handling around individual request processing so an error will not stop the entire process.


rfox2:
:+1:

jgondron:
:+1:

PR #116, opened 2015-09-02
AIMS-323 Replace non-utf8-encoded characters in requests

jaronkk:
Currently a non-utf8 character in the request data will cause the sync task to error and not update the request.

This calls encode on all strings in the request data to ensure there are no invalid characters during save.


rfox2:
:+1:  Thanks Jaron

jgondron:
:+1:

PR #117, opened 2015-09-03
AIMS-321: Batch items filter

jgondron:
Why: User would like to additionally be able to filter the items that match a request.
How: Each sub table will now use DataTables. I also had to modify the way that the global request filters work since these are applied to all datatables, not just specific instances.


rfox2:
:+1:

jaronkk:
@rfox2 Can you re-review this with the new changes?

jaronkk:
:+1: For @jgondron's  stuff, and he already :+1:ed https://github.com/ndlib/annex-ims/pull/118

rfox2:
:+1:

PR #118, opened 2015-09-03
Batch items results list styles and enhancements

jaronkk:
I perhaps went overboard in exploring this...

1. Indent the item list row slightly
2. Make sure the child row has the same background color as the parent row
3. Left align the search box
4. Conditionally show the search box and pagination if there are enough records.
5. Default to sorting by chron, added some hopefully intelligent sorting on chron so it sorts numerically rather than by string value

Example with filtering:
annex_ims

Example without filtering:
annex_ims

### DIFF
--- a/app/assets/stylesheets/batches.css.scss
+++ b/app/assets/stylesheets/batches.css.scss
@@ -1,0 +1,21 @@
+.table > tbody > tr.batch-request-results-row > td {
+  border-top-style: dashed;
+}
+
+.batch-request-results {
+  border-left: 1px solid #ddd;
+  padding: 0 0 1em 2em;
+
+  .dataTables_filter {
+    float: left;
+    margin-right: 1em;
+
+    label {
+      float: none;
+    }
+  }
+
+  .dataTables_length {
+    float: left;
+  }
+}
--- a/app/views/batches/index.html.haml
+++ b/app/views/batches/index.html.haml
@@ -133,9 +133,16 @@
         }
     } );
 
-    $("#selectAll").click(function() {
-      $(".item").prop("checked", $("#selectAll").prop("checked"));
+    $("#selectAll").click(function(event) {
+      event.stopPropagation();
+      $(".item").prop("checked", $("#selectAll").prop("checked")).each(function() {
+        checkboxChanged(this);
+      });
     } );
+
+    $("form").on("change", ".item", function() {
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/batches.css.scss
```diff
@@ -0,0 +1,21 @@
...
+  .dataTables_filter {
+    float: left;
+    margin-right: 1em;
+
+    label {
+      float: none;
+    }
+  }
+
+  .dataTables_length {
```
jaronkk: This class is created by datatables, can't change it.

app/assets/stylesheets/batches.css.scss
```diff
@@ -0,0 +1,21 @@
+.table > tbody > tr.batch-request-results-row > td {
+  border-top-style: dashed;
+}
+
+.batch-request-results {
+  border-left: 1px solid #ddd;
+  padding: 0 0 1em 2em;
+
+  .dataTables_filter {
```
jaronkk: This class is created by datatables, can't change it.

app/assets/stylesheets/batches.css.scss
```diff
@@ -0,0 +1,21 @@
+.table > tbody > tr.batch-request-results-row > td {
```
jaronkk: Needs to be formatted this way to override bootstrap styles.

app/views/batches/index.html.haml
```diff
@@ -167,17 +176,48 @@
...
+          targets: resultsColumns.chronSort,
+          searchable: false,
+          visible: false,
+        },
+      ],
+    }
+
+    if (tr.data('items').length > 10) {
+      options.searching = true;
+      options.paging = true;
```
jgondron: Can you test this a bit more? If I select two items that are on two different pages, it doesn't seem to add them both to the matches table. If I select two items on the current page, it works fine.

jaronkk: Ah, good point, I'll see what I can do with that

jaronkk: @jgondron Since datatables moves row elements out of the dom when it's not displaying them, I had to add a hidden form field for each checkbox that is checked.  When the checkbox is checked, the input element is created, and when it's unchecked, the input element is removed.

jgondron:
:+1:

PR #119, opened 2015-09-03
AIMS-324 configurable item search per page

jaronkk:
When we increased the hardcoded limit in SearchItems to 5,000, it increased the number of results being displayed on the item search page to that amount.

This reverts the default back to 50, and updates the BuildRequestData service to request 5,000 per page.

I also added a spec for the SearchItems service so we have complete coverage on that file.


jgondron:
:+1:

rfox2:
:+1:

PR #120, opened 2015-09-08
AIMS-325: Session timeout

jgondron:
Why: User sessions remain valid for indefinite amount of time, causing security issues.
How:  User sessions will now expire. Did not use built in devise timeoutable method since it doesn't allow customizing the timed out page. Created our own expiration calculation which will now render a timed out page that forces the user to click login. By default, will timeout after 10 hours of inactivity for prod/preprod/staging. This should cause a login once per day if they're active in it all day. For dev using 24 hours. These can be changed in configs.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   before_action :check_authentication
+  before_action :check_activity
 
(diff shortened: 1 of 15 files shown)


rfox2:
:+1:

rfox2:
Proxy for Jaron :+1:

PR #121, opened 2015-09-09
AIMS-325: Session timeout

jgondron:
Discovered we still had a problem with it retaining the previous app session even with the renew call. Added a sign_out call and it seems to now be fully functional.


jaronkk:
:+1:

rfox2:
:+1:

PR #123, opened 2015-10-29
Aims 334 shelf transfers

rfox2:
What: Added functionality so that it is easier to do full shelf transfers where all trays from one shelf are being transferred to another shelf.
How: Added a transfer model, several UI components along with ancillary functionality


app/assets/stylesheets/framework_and_overrides.css.scss
```diff
@@ -175,6 +175,10 @@ a.list-group-item[aria-expanded="true"] {
     background-image: image_url("traytoshelf.png");
     padding-left: 95px;
   }
+  .shelft {
+    background-image: image_url("shelftoshelf.png");
```
rfox2: This is actually a method from rails so we can't make this change

app/controllers/transfers_controller.rb
```diff
@@ -0,0 +1,89 @@
...
+    tray_barcode = @tray.barcode
+
+    if @shelf.blank?
+      flash[:error] = "Shelf with barcode #{shelf_barcode} does not exist. Rescan tray."
+      redirect_to transfer_path(id: params[:id])
+      return
+    end
+
+    if dissociate_and_reassociate_tray(@tray, @shelf, @transfer.shelf)
+      flash[:notice] = "Tray with barcode #{tray_barcode} transfered."
```
jgondron: Add which shelf it was transferred to

app/controllers/transfers_controller.rb
```diff
@@ -0,0 +1,95 @@
...
+  def view_active
+    @transfers = Transfer.all
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def scan_tray
+    @transfer = Transfer.find(params[:id])
+    tray_barcode = params[:tray][:barcode]
+    @tray = Tray.where(barcode: tray_barcode).take
+    @shelf = Shelf.new
```
jgondron: Is this needed by anything?

rfox2: You mean the class variable?

app/assets/stylesheets/framework_and_overrides.css.scss
```diff
@@ -166,19 +166,28 @@ a.list-group-item[aria-expanded="true"] {
...
 
   }
+
   .tray {
     background-image: image_url("traytoshelf.png");
     padding-left: 95px;
   }
+
+  .shelft {
+    background-image: image_url("shelftoshelf.png");
```
rfox2: I think we can safely ignore this.

ericleasemorgan: Ignore it, yes, but can it be turned off? There are many! --E

jgondron:
Need specs for the new controller and the new activity logger method

jgondron:
:+1: Once coverage is fixed

PR #125, opened 2015-11-06
AIMS-334 Bug fix and add activity logger for destroy shelf

rfox2:
What:  Fixed a bug with the shelf parameter reference, and added new method for logging transfer destroy actions

### DIFF
--- a/app/controllers/transfers_controller.rb
+++ b/app/controllers/transfers_controller.rb
@@ -86,7 +86,7 @@ class TransfersController < ApplicationController
 
   def check_for_blank_shelf(type)
     if @shelf.blank?
-      flash[:error] = "Shelf with barcode #{params[:shelf][:barcode]} does not exist. Please rescan."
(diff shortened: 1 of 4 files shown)


jonhartzler:
:+1:

app/controllers/transfers_controller.rb
```diff
@@ -101,6 +101,7 @@ def check_for_blank_shelf(type)
   def check_for_final_tray
     if @transfer.shelf.trays.count == 0
       @transfer.destroy!
```
jgondron: We should probably use a service class for destroying a transfer and put the destroy and log in there.

rfox2: The reason I'd say that isn't necessary is because elsewhere in the app, the activity logger is directly called. This is the only circumstance where a transfer would be deleted. I did think of that, but it seems like overkill.

jgondron:
:+1:

PR #126, opened 2015-11-09
AIMS-334 Add function to delete transfers, and add date/time to active transfer list

rfox2:
What: Added service class for destroying transfers, added controller method for doing so, and added a date / time stamp to the active transfer list

### DIFF
--- a/app/controllers/transfers_controller.rb
+++ b/app/controllers/transfers_controller.rb
@@ -45,7 +45,7 @@ class TransfersController < ApplicationController
     if DestroyTransfer.call(Transfer.find(params[:id]), current_user)
       flash[:notice] = "Transfer canceled."
     else
-      flash[:error] = "Transer not canceled due to system error."
+      flash[:error] = "System Error: Transaction has been canceled. Please try again and/or open a support ticket."
     end
     redirect_to view_active_transfers_path
   end
--- a/app/controllers/transfers_controller.rb
+++ b/app/controllers/transfers_controller.rb
@@ -42,10 +42,11 @@ class TransfersController < ApplicationController
   end
 
   def destroy
-    if DestroyTransfer.call(Transfer.find(params[:id]), current_user)
+    returned_value = DestroyTransfer.call(Transfer.find(params[:id]), current_user)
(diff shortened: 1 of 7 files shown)


app/controllers/transfers_controller.rb
```diff
@@ -41,6 +41,15 @@ def create
     redirect_to transfer_path(id: transfer.id)
   end
 
+  def destroy
+    if DestroyTransfer.call(Transfer.find(params[:id]), current_user)
+      flash[:notice] = "Transfer canceled."
+    else
+      flash[:error] = "Transer not canceled due to system error."
```
jonhartzler: Perhaps: 

System Error.  The transaction has not been canceled.   Please try again and if it is unsuccessful --insert text about what they can do.

app/services/destroy_request.rb
```diff
@@ -15,4 +15,12 @@ def destroy
     ActivityLogger.remove_request(request: request, user: user)
     status
   end
+
+  def self.shelve_trays(shelf, user)
```
jonhartzler: Should this method have spec?

app/controllers/transfers_controller.rb
rfox2: I'll rewrite this message

app/services/destroy_request.rb
rfox2: Yes, probably. I'll add it.

app/services/destroy_transfer.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def initialize(transfer, user)
+    @transfer = transfer
+    @shelf = transfer.shelf
+    @user = user
+  end
+
+  def destroy
+    status = @transfer.destroy!
```
jgondron: Can we wrap this in a transaction block? If anything happens with reshelving any of the trays, or with destroying the transfer, we would end up in a partial state.

rfox2: Done

jgondron:
:+1:

jonhartzler:
:+1:

PR #129, opened 2016-01-05
AIMS-331:  Empty batch bug

jgondron:
Why: If multiple scan requests for a single item would get batched, the user would only be able to process one of them.
How: The application was immediately dissociating an item from a bin as soon as the first scan was processed, which prevented the view from showing the remaining matches. Changed the app to only dissociate an item from a bin once all requests in that bin for that item have been fulfilled. Now, when they process one request, it will flash a message warning that there are remaining requests for that item. In addition, I created a rake task that we can use to spot check and clean up the data when we deploy.


rfox2:
:+1:

PR #130, opened 2016-01-06
AIMS-336: Add process details

jgondron:
Why: Now that they are able to process requests directly in the Annex, they need more information on the request to be displayed when processing a bin.
How: Added Article Title and Description. Description is a packed field that contains things like pages, volume, etc. We also decided to change "Process Batch" to "Process Bin". The view linked to by "Process Batch" is actually showing a list of bins. The desired behavior for the workflow is to show all bins, not just the bins associated with that user's current batch. So we're renaming the menu option to match so that there's less confusion for future users/devs.

### DIFF
--- a/app/views/bins/show.html.haml
+++ b/app/views/bins/show.html.haml
@@ -11,6 +11,8 @@
         %th= 'Transaction'
         %th= 'Title'
         %th= 'Author'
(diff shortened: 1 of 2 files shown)


rfox2:
:+1:

PR #131, opened 2016-01-07
AIMS-219: Cleanup "not found" items that are not associated with a tray

jgondron:
Why: A lot of items have been added to the system using junk data from the barcode scanner due to the way we optimistically create an item, then look up its info later. This is getting in the way of querying/viewing valid items.
How: Created a rake task to destroy all items that are not assigned to a tray and are either "not found" or "not marked for annex". This task will be run adhoc to clean up the data (such as during our next deploy). Future development such as input validation should begin to prevent this data from being created, so a cron job is likely not needed, but if we do end up creating one, it can just use this rake task.

### DIFF
--- a/app/services/destroy_item.rb
+++ b/app/services/destroy_item.rb
@@ -2,7 +2,7 @@ class DestroyItem
   attr_reader :item, :user
 
   def self.call(item, user)
-    new(item, user).destroy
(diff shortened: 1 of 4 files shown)


rfox2:
:-1:

rfox2:
:+1:

PR #132, opened 2016-01-11
AIMS-339 Close out batch with last match

rfox2:
What: Added functionality that handles the logic of match deletion. If the last match from a batch is removed, the batch itself will be closed out.
Why: If the last match from a batch is deleted, it leaves the batch in an undetermined state from which the user cannot extract it. We are making a change to the IMS that will prevent the creation of new batches if an existing batch is active, so this is a necessary prerequisite.
How: Created a service class for deleting matches, and added some controller logic that takes advantage of the new class. The remove method of the batches controller was heavily refactored.


app/controllers/batches_controller.rb
```diff
@@ -34,19 +34,12 @@ def current
...
-    if !params[:match_id].blank?
-      match = Match.find(params[:match_id])
-      if !match.blank?
-        ActivityLogger.remove_match(item: match.item, request: match.request, user: current_user)
-        match.destroy!
-      end
-    end
-
-    redirect_to current_batch_path
+    check_for_params(params[:match_id])
```
jgondron: Are you intending for this to stop execution here when this param is blank? Currently, it will tell rails to redirect, but it will still continue to the DestroyMatch call which will fail. 

It seems like we could just remove this call and the check_for_params method. If ever we got to this point when the match_id is nil, then something went wrong and we should get an errbit notification about the exception.

app/services/destroy_match.rb
```diff
@@ -0,0 +1,33 @@
...
+    ActiveRecord::Base.transaction do
+      @match.destroy!
+      ActivityLogger.remove_match(item: @match.item, request: @match.request, user: @user)
+      determine_batch_status
+    end
+  rescue StandardError => e
+    e.message
+  end
+
+  def determine_batch_status
```
jgondron: Could we just return a boolean here? You only have two possible outcomes, seems like a boolean would be sufficient to notify the caller of success/failure

app/services/finish_batch.rb
```diff
@@ -11,11 +11,10 @@ def initialize(batch)
 
   def finish!
     @batch.skipped_matches.each do |s|
-      s.destroy! #accepted matches will remain, so those requests will stay out of queue. Skipped requests will go back in the queue.
+      s.destroy! # accepted matches will remain, so those requests will stay out of queue. Skipped requests will go back in the queue.
```
jgondron: Should this use the DestroyMatch service?

app/services/complete_batch.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize(batch, user)
+    @batch = batch
+    @user = user
+  end
+
+  def complete_batch!
+    FinishBatch.call(@batch, @user) unless remaining_matches?
+  end
+
+  def remaining_matches?
```
jgondron: Seems like this conditional check could be moved into FinishBatch. Which brings up another question: Should this only check for unskipped matches? Looks like if a batch had skipped matches, we would never be able to complete it.

jgondron:
:+1: Once the comment is addressed

PR #133, opened 2016-01-12
AIMS-329 Add logic for batch creation / completion

rfox2:
What: Added some logic so that a user can only have one active batch at a time
Why: It was possible previously for a user to have multiple batches running at the same time and this can lead to confusion and errors. We wanted the user to only be able to initiate a single batch at a time.
How: Added some functionality to the batch controller methods as well as a couple of private methods. 
This pull request is related to and dependent on https://github.com/ndlib/annex-ims/pull/132


app/controllers/batches_controller.rb
```diff
@@ -211,4 +218,27 @@ def cancel_single_active
 
     redirect_to view_active_batches_path
   end
+
+  private
+
+  def flash_message(type)
```
jgondron: Should probably just use i18n for this

jgondron:
:+1: Once coverage and comment is addressed

PR #134, opened 2016-01-12
AIMS-263: Remove item from bin

jgondron:
Why: Currently, there is no way to remove an item from a bin once it's added without completing the request. This means if the incorrect item was pulled, there is no way to correct it.
How: Added a Remove button to the bin view and a remove action to the controller so that the user can remove a match from a bin. Refactored match destroying a bit. We have more than one thing that needs to destroy matches, so moved this into a DestroyMatch service. Changed all direct calls to direct match.destroy to use the new service, except for within CancelBatch. After speaking with Miranda, this should never occur since we should not allow canceling a batch once a match is made (created AIMS-341)


app/controllers/bins_controller.rb
```diff
@@ -7,7 +7,21 @@ def show
...
+    DestroyMatch.call(match: @match, user: current_user)
+
+    unless bin.matches.where(item: item).empty?
+      flash[:warning] = "Removed transaction #{@match.request.trans}. There are remaining requests for the item."
+    end
+
+    redirect_to show_bin_path(id: bin.id)
+  end
+
+  def process_match
```
jgondron: This is an outdated message

app/services/destroy_match.rb
```diff
@@ -0,0 +1,40 @@
...
+    false
+  end
+
+  private
+
+  def item
+    match.item
+  end
+
+  def can_remove_item?
```
rfox2: I've got a method that does the same thing - determine_batch_status. I can change the code for that method so that it returns a boolean value but that also has a method call to complete a batch if the last item is removed.

app/services/destroy_match.rb
```diff
@@ -0,0 +1,40 @@
...
+    match.bin.matches.where(item: item).empty?
+  end
+
+  def dissociate_bin
```
rfox2: We should merge this into a common DestroyMatch class, but also roll it into the overall process.

rfox2:
:+1:

PR #136, opened 2016-01-15
AIMS-263: Remove item from bin

jgondron:
Changing appearance of the remove button. Will now say Skip and look like the Skip button used in batch retrieval. It's a similar operation that the user is familiar with, so they should better understand what this action will do at this point in the workflow.


rfox2:
:+1:

PR #137, opened 2016-01-15
Aims 340 add bin info to item details

rfox2:
What: Added information to the item detail page that indicates the current location of the item.
Why: It had been difficult to determine the current status of an item if it had been retrieved for batch processing and was sitting in a bin awaiting processing, having been shipped, or having returned from a loan awaiting restocking. This information will help to alleviate that problem.
How: Created a new method in the item view presenter that returns location strings and added that information to the item detail view.


jgondron:
:+1:

PR #142, opened 2016-02-04
AIMS-335: Add worker permissions

BiigY:
Hi there,

I need a review of my code that updated some files in Annex project. Currently, I am still working on other spec unit testing. So as soon as I finish them, I will push these changes as well. For now, I would like to see any feedback and suggestions.

Thanks!

### DIFF
--- a/app/controllers/batches_controller.rb
+++ b/app/controllers/batches_controller.rb
@@ -1,5 +1,8 @@
 class BatchesController < ApplicationController
   def index
+    # Only admin has rights to call this method
+    require_admin
+
     # Should this be in a service object? It's a relatively simple one-liner.
     requests = Request.all.where("id NOT IN (SELECT request_id FROM matches)")
     @data = BuildRequestData.call(requests)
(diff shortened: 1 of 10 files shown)


app/controllers/batches_controller.rb
```diff
@@ -1,5 +1,8 @@
 class BatchesController < ApplicationController
   def index
+    # Only admin has rights to call this method
+    require_admin
```
jgondron: Since we are still doing a fairly simple type of access control, and are basically saying everything in batches is exclusive to admins, perhaps we should DRY this up by using a before_action instead of duplicating require_admin in each action

app/views/users/index.html.haml
```diff
@@ -1,36 +1,39 @@
...
-    %thead
-      %tr
-        %th= 'User'
-        %th= 'Sign in Count'
-        %th= 'Last Sign In'
-        %th= 'Admin?'
-        %th
-    %tbody
-      - @users.each do |user|
+- if user_admin?
```
jgondron: Should use require_admin in the controller similar to the batches controller instead of checking this in the view.

jgondron:
Rename the branch and add a prefix of AIMS-335-. Ex: AIMS-335-Add-worker-permissions. This is for traceability between tickets and PRs.

PR #145, opened 2016-02-16
Aims 335 add worker permissions

BiigY:
I need a review of my code, to be more exact, I updated some controller files in Annex project, as well as wrote some spec unit testing. All changes were pushed already. I would like to see any feedback and suggestions.

### DIFF
--- a/app/views/users/index.html.haml
+++ b/app/views/users/index.html.haml
@@ -34,5 +34,3 @@
           window.table = $('#users').DataTable();
           $('div.dataTables_filter input').focus();
         } );
-- else
-- raise_404
--- a/app/views/users/index.html.haml
+++ b/app/views/users/index.html.haml
@@ -5,7 +5,8 @@
   = form_tag create_users_path, method: "post" do |f|
     %div{ :style => "display: inline-block; width: 100%;" }
       %div{ :style => "float: left;" }
-        %input{ name: "user_name", id: "user_name", type: "text", required: true }/
+        %input{ name: "user_name", id: "user_name", type: "text",
+        required: true }/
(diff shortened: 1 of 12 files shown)


app/views/users/index.html.haml
```diff
@@ -34,3 +34,5 @@
           window.table = $('#users').DataTable();
           $('div.dataTables_filter input').focus();
         } );
+- else
```
jgondron: Is this left over from the if admin check that you removed?

config/initializers/devise.rb
```diff
@@ -4,7 +4,7 @@
   # The secret key used by Devise. Devise uses this key to generate
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
-  # config.secret_key = '011b6001f75749566e092721113095fc81cef7c845b0854f0e3280861558b976744c6d080c9f921a7a07bc251d1a07ce558d425ce92b80f1e5ec3e519b001f7d'
+  config.secret_key = "011b6001f75749566e092721113095fc81cef7c845b0854f0e3280861558b976744c6d080c9f921a7a07bc251d1a07ce558d425ce92b80f1e5ec3e519b001f7d"
```
jgondron: Can you double check to see if this was necessary?

jgondron:
:+1:  after hound comments are addressed

rfox2:
This looks good Yerlan! The hound comments need to be addressed - those are code style issues that we try to adhere to. After that, :+1: and you can merge this with master. Make sure the Jira ticket then gets moved over into QA.

PR #146, opened 2016-03-29
Updates for users page AIMS-335

BiigY:
Admins can now update the type of user for existing users in the database via the users page, which now has select options for user types and update button. Every time the user type is changed, there will be a notification on top of the page that will give details of the update. Also the current user is disabled to change own user type due to a safety requirement.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -17,12 +17,22 @@ class UsersController < ApplicationController
 
   def update
     user_id = params["user_id"]
-    # If the checkbox was false on submit, the admin param will be null :(
-    admin = params["admin"] || false
     user = User.find(user_id)
     if user
-      user.admin = admin
+      # If the admin was selected, then user is set to admin and otherwise to worker
+      if params["user_type"] == "admin"
+        user.admin = true
+        user.worker = false
+      # If the worker was selected, then user is set to worker and otherwise to disabled
+      elsif params["user_type"] == "worker"
+        user.admin = false
+        user.worker = true
+      else
+        user.admin = false
+        user.worker = false
+      end
       user.save!
+      flash[:notice] = "#{user.username}'s status is updated to #{params["user_type"]}."
     end
     redirect_to users_path
   end
--- a/app/views/users/index.html.haml
+++ b/app/views/users/index.html.haml
@@ -26,16 +26,12 @@
           %td= user.last_sign_in_at
 
           - if current_user == user
-            %td= 'admin'
+            %td= GetUserType.call(user)
             %td
           - else
             = form_tag update_users_path, method: "put" do |f|
-              - if user.admin == true
-                %td= select_tag :user_type, options_for_select(["admin", "worker", "disabled"], "admin")
-              - elsif user.worker == true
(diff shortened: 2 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -17,12 +17,22 @@ def create
...
+      # If the worker was selected, then user is set to worker and otherwise to disabled
+      elsif params["user_type"] == "worker"
+        user.admin = false
+        user.worker = true
+      else
+        user.admin = false
+        user.worker = false
+      end
       user.save!
+      flash[:notice] = "#{user.username}'s status is updated to #{params["user_type"]}."
```
rfox2: This will need to be corrected.

app/views/users/index.html.haml
```diff
@@ -24,12 +24,22 @@
...
+              - if user.admin == true
+                %td= select_tag :user_type, options_for_select(["admin", "worker", "disabled"], "admin")
+              - elsif user.worker == true
+                %td= select_tag :user_type, options_for_select(["admin", "worker", "disabled"], "worker")
+              - else
+                %td= select_tag :user_type, options_for_select(["admin", "worker", "disabled"], "disabled")
+              %td
+                = hidden_field_tag :user_id, user.id
+                .actions
+                  = submit_tag 'Update', class: 'btn btn-primary', id: "user-#{user.id}"
```
rfox2: I would ignore these hound warnings. This is fine.

app/controllers/users_controller.rb
```diff
@@ -17,12 +17,22 @@ def create
 
   def update
     user_id = params["user_id"]
-    # If the checkbox was false on submit, the admin param will be null :(
-    admin = params["admin"] || false
     user = User.find(user_id)
     if user
-      user.admin = admin
+      # If the admin was selected, then user is set to admin and otherwise to worker
+      if params["user_type"] == "admin"
```
rfox2: Where does your code prevent the logged in user from turning their own admin flag off? If they try to do that we:

1. Want to prevent that from happening
2. Want to send a warning message via flash that they tried to take themselves off of the admin list

spec/controllers/users_controller_spec.rb
```diff
@@ -8,18 +8,27 @@
   end
 
   describe "PUT update" do
```
rfox2: Add a test that verifies that the currently logged in user cannot remove themselves from the admin list.

rfox2:
Yerlan, I added some comments. Can you check?

spec/controllers/users_controller_spec.rb
```diff
@@ -8,18 +8,27 @@
...
-    it "sets admin to the value given" do
+    it "sets admin to the value given for user type" do
       expect_any_instance_of(User).to receive(:admin=).with(true)
-      put :update, user_id: 1, admin: true
+      expect_any_instance_of(User).to receive(:worker=).with(false)
+      put :update, user_id: 1, user_type: "admin"
     end
 
-    it "sets admin to the value given" do
+    it "sets admin to the value given for user type" do
```
jgondron: Use different descriptions for these tests. Ex: "sets admin and worker properties to the correct values for user type 'worker'"

rfox2:
After you've satisfied Justin's concerns, :+1:

PR #147, opened 2016-04-05
Aims 345 item validation parameters

BiigY:
Completed all ticket requirements with test specs.


app/controllers/trays_controller.rb
```diff
@@ -228,4 +192,66 @@ def withdraw
...
   end
+
+  def invalid
+    @tray = Tray.find(params[:id])
+    @thickness = params[:thickness]
+    @barcode = params[:barcode]
+    @id = params[:user_id]
+  end
+
+  def create_item(tray=Tray.find(params[:id]), barcode=params[:barcode], id=params[:user_id], thickness=params[:thickness])
```
rfox2: Spaces need to be added between the elements of the parameter list. Ignore the ABC warning.

app/controllers/trays_controller.rb
```diff
@@ -131,71 +131,35 @@ def show_item
...
     @barcode = params[:barcode]
+    @thickness = params[:thickness]
   end
 
   def associate_item
     @tray = Tray.find(params[:id])
     @size = TraySize.call(@tray.barcode)
 
     barcode = params[:barcode]
+    thickness = params[:thickness]
```
rfox2: Since an instance variable is being assigned on line 134, do you need this assignment here? Why not just refer to @thinkness?

app/models/issue.rb
```diff
@@ -1,5 +1,5 @@
 class Issue < ActiveRecord::Base
-  ISSUE_TYPES = ["not_found", "not_for_annex", "aleph_error"]
+  ISSUE_TYPES = ["not_found", "not_for_annex", "aleph_error", "not_valid_barcode"]
```
rfox2: Just add .freeze to the end of this array.

app/services/is_valid_item.rb
```diff
@@ -0,0 +1,23 @@
...
+    # If there is more than one regular expression set, the barcode is valid if it matches any one of them.
+    parameters.each do |parameter|
+      if parameter['enabled']
+        if barcode =~ /^#{parameter['regex']}(.*)/
+          return true
+        end
+      end
+    end
+
+    return false
```
rfox2: You don't need "return" here. Just the boolean "false". It will automatically get returned if the method reaches this point.

spec/features/trays_spec.rb
```diff
@@ -294,6 +294,18 @@
       expect(page).to have_content 'select a valid thickness'
     end
 
+    it "requires a valid width for an item" do
```
rfox2: I think you mean thickness here. Not width.

app/controllers/trays_controller.rb
```diff
@@ -228,4 +192,66 @@ def withdraw
...
+    @tray = Tray.find(params[:id])
+    @thickness = params[:thickness]
+    @barcode = params[:barcode]
+    @id = params[:user_id]
+  end
+
+  def create_item(tray=Tray.find(params[:id]), barcode=params[:barcode], id=params[:user_id], thickness=params[:thickness])
+    begin
+      item = GetItemFromBarcode.call(barcode: barcode, user_id: id)
+    rescue StandardError
```
rfox2: My reservation about capturing the exception this way is that it will capture ALL exceptions, and other valid exceptions may go unnoticed. I'm wondering if we shouldn't make this rescue more specific to the type of exception we really want to trap and let the other bubble up.

app/controllers/trays_controller.rb
```diff
@@ -228,4 +192,66 @@ def withdraw
...
+      flash[:error] = I18n.t("errors.barcode_not_found", barcode: barcode)
+      redirect_to show_tray_item_path(id: tray.id, barcode: barcode, thickness: thickness)
+      return
+    end
+
+    # Only if the item barcode wasn't found in the database and was set aside by the system
+    if item.nil? && params[:user_id].nil?
```
rfox2: Why test for the user id at this point?

app/controllers/trays_controller.rb
```diff
@@ -228,4 +192,66 @@ def withdraw
...
+      flash[:error] = I18n.t("errors.barcode_not_found", barcode: barcode, thickness: thickness)
+      redirect_to missing_tray_item_path(id: tray.id)
+      return
+    # When the item barcode wasn't found in the database and was set aside by the user
+    elsif !params[:user_id].nil?
```
rfox2: Similar to my other question: I'm not following why user_id is tested as a condition for barcode validity...?

app/controllers/trays_controller.rb
```diff
@@ -228,4 +192,66 @@ def withdraw
...
+        flash[:notice] = "Item #{barcode} already assigned to #{tray.barcode}. Record updated."
+      else
+        flash[:notice] = "Item #{barcode} stocked in #{tray.barcode}."
+      end
+      if TrayFull.call(tray)
+        flash[:error] = "warning - tray may be full"
+      end
+      redirect_to show_tray_item_path(id: tray.id)
+      return
+    rescue StandardError => e
```
rfox2: Since all errors will be trapped here, and it doesn't interrupt application execution, should we send a flash message to the user indicating that there was a problem?

app/services/is_valid_item.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    #  If there is no regular expression set, the system will skip validation.
+    if parameters.count == 0
+      return true
+    end
+
+    # If there is more than one regular expression set, the barcode is valid if it matches any one of them.
+    parameters.each do |parameter|
+      if parameter["enabled"]
+        if barcode =~ /^#{parameter["regex"]}(.*)/
```
rfox2: I'm sure there's a good reason for this, but why all the trouble to create an external parameter for this test? Is there a reason why we don't want to hard code the regex into this method? (I know - "magic numbers") ... ?

rfox2:
Yerlan: I have some questions about the code and I'd like Justin to take a look at parts of the code as well before we greenlight this for QA. Can you take a look at my questions?

app/controllers/trays_controller.rb
```diff
@@ -131,71 +131,35 @@ def show_item
...
 
-    begin
-      item = GetItemFromBarcode.call(barcode: barcode, user_id: current_user.id)
-    rescue StandardError
-      flash[:error] = I18n.t("errors.barcode_not_found", barcode: barcode)
-      redirect_to show_tray_item_path(id: @tray.id, barcode: barcode, thickness: params[:thickness])
+    # The system should validate the barcode against the stored regular expression(s)
+    unless IsValidItem.call(barcode)
+      flash[:error] = I18n.t("errors.barcode_not_valid", barcode: barcode)
+      redirect_to invalid_tray_item_path(id: @tray.id, thickness: thickness, barcode: barcode, user_id: current_user.id)
```
jgondron: Seems like a redirect is not necessary here. From what I can tell, you're just redirecting in order to render a different view. You could just render that view directly here instead.

app/controllers/trays_controller.rb
```diff
@@ -131,71 +131,35 @@ def show_item
...
-      end
-      redirect_to show_tray_item_path(id: @tray.id)
-      return
-    rescue StandardError => e
-      notify_airbrake(e)
-      flash[:error] = e.message
-      redirect_to show_tray_item_path(id: @tray.id)
-      return
-    end
+    create_item(@tray, barcode, current_user.id, thickness)
```
jgondron: Creation of an item would be better handled in service object. You would then need to figure out a way for the controller to act on the results of that call. So perhaps the service object returns an error when it fails, and the controller can decide what to render based on that error.

config/parameters.yml
```diff
@@ -0,0 +1,27 @@
+# Store bar-code validation parameters here. Acessible from Rails.configuration.parameters
+# Each parameter will be written as a regular expression
+# It needs to be possible to store multiple parameters
+# The parameters need to be stored in such a way that a UI can be built later to manage them
+# It should be possible to "inactivate" a parameter without deleting it (some sort of flag)
+
+common: &defaults
+  # The initial barcode parameter should be: 14 digits with no additional
+  # characters before or after. Flag (in/de)activates it without deleting it.
+  parameters:
```
jgondron: Not a fan of the naming of this. Call it something more appropriate, like barcode_patterns, or something more fitting than just parameters.

app/services/is_valid_item.rb
jgondron: @rfox2 They needed the ability to expand the options in the future, so we decided to make it configurable.

app/controllers/trays_controller.rb
BiigY: Initially it was there, however, since @thickness is defined under in a different method, it cannot be accessed in associate_item method. Also, some specs fail if I use @thickness.

spec/features/trays_spec.rb
BiigY: Good catch, I was following the previous spec, which used the width instead of thickness. However, I updated both of them.

app/controllers/trays_controller.rb
BiigY: Can we create a new ticket for this suggestion?

app/controllers/trays_controller.rb
BiigY: As the comment mentions, we need to know whether the request to set aside an item was done by the user or the system.

app/controllers/trays_controller.rb
BiigY: I tested it out that the only way to detect that the actual user requested to set aside the item is to check if the user parameter is passed by and included in the request.

rfox2:
👍

jgondron:
👍

PR #150, opened 2016-04-26
AIMS-344: Remove button is displayed only for items that have no bin barcode.

BiigY:
If the item is added to the bin, then that item will have bin barcode presented instead of Remove button in View Current Batch page.


jgondron:
👍

rfox2:
👍

PR #151, opened 2016-05-04
AIMS-354: Added validation for count of items in the tray

BiigY:
When a tray is full the user clicks "Done".  At that point the user is prompted to "Enter the number of items in the tray."
If the count matches, the user is taken to the Stocking page to scan the next Tray.
If the count does NOT match, the user receives an error message and is prompted to recount, and enter the count again.

### DIFF
--- a/app/controllers/trays_controller.rb
+++ b/app/controllers/trays_controller.rb
@@ -233,7 +233,7 @@ class TraysController < ApplicationController
 
     if !tray_count.nil?
       if tray_count.to_i != items.count
-        flash.now[:error] = "Manual and System Tray counts don't match! Try again."
+        flash.now[:error] = I18n.t("trays.items_count_not_match")
       else
         redirect_to trays_items_path
       end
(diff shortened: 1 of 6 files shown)


app/controllers/trays_controller.rb
```diff
@@ -225,4 +225,18 @@ def create_item(tray = Tray.find(params[:id]), barcode = params[:barcode], thick
       redirect_to show_tray_item_path(id: tray.id)
     end
   end
+
+  def count_item
```
rfox2: Let's name this item_count instead - or count_items just so that the meaning is clearer

rfox2:
I made one comment but I have a couple of questions. When the count is entered, the user should have two chances to get the count correct before the tray is put on the "trays with issues" list. Is that functionality intended to be here or in another PR? Also, Miranda may not have mentioned this but this also needs to be tied into the activity log. So, for the transition from the initial count to being put onto the tray issue list will need an activity log. There will also need to be an activity log for any transitions that take place when the tray is being worked on from the "trays with issues" list - for example if the count problem is corrected and the tray then goes back into the standard workflow.

PR #152, opened 2016-05-05
AIMS-290: Add validation to item thickness field

BiigY:
Added extra validation for thickness and updated fields to require values


rfox2:
👍

jgondron:
👍

PR #155, opened 2016-05-17
AIMS-355: speeds up counting items in tray.

BiigY:
The way of calculating items in the tray was updated to more efficient way.
Also the indication of a number of items from show_item page removed.


jgondron:
👍

PR #157, opened 2016-06-21
Aims 356 add trays with issues

rfox2:
What: Added a workflow doing quality assurance against trays, in order to validate the count of items the worker thinks is in the tray vs. what the system has recorded for that tray. This also includes the resolution process including two new reports and the ability to resolve tray issues by submitting the correct tray / item count.
Why: Trying to avoid some of the QA issues we've had with trays being shelved having missing or misplaced items.
How: Refactored the trays controller; Added several service classes; Added new logger methods, Added query object; Added several views for reporting purposes; Modified specs;


app/views/trays/tray_detail.html.haml
```diff
@@ -1,5 +1,47 @@
...
+          %tr
+            %th= "Activity"
+            %th= "When"
+            %th= "Shelf"
+            %th= "Who"
+        %tbody
+          - @history.each do |record|
+            %tr
+              %td= t("activity_log." + record.action)
+              %td= record.action_timestamp.strftime("%m-%d-%Y %I:%M%p")
```
rfox2: Ignoring this

app/controllers/trays_controller.rb
```diff
@@ -239,14 +239,51 @@ def count_items
...
           @validation_count_items = @validation_count_items.to_i + 1
           if @validation_count_items == 2
+            AddTrayIssue.call(user: current_user, tray: @tray, message: "Tray count invalid", type: "incorrect_count")
             flash.now[:error] = I18n.t("trays.count_validation_not_pass")
           else
             flash.now[:error] = I18n.t("trays.count_items_not_match")
           end
         end
       else
+        if invalid_count_issues?
```
jgondron: Does this mean that you can resolve an issue just by going through this count process again? I thought a user would have to go through the Trays with Issues view to resolve all issues.

app/controllers/trays_controller.rb
```diff
@@ -239,14 +239,51 @@ def count_items
...
+  def tray_detail
+    @tray = Tray.where(barcode: params[:barcode]).take
+    if @tray
+      @history = ActivityLogQuery.tray_history(@tray)
+    end
+  end
+
+  private
+
+  def invalid_count_issues?
```
jgondron: This should be encapsulated within a query object, maybe an alternate method within IssuesForTrayQuery

app/controllers/trays_controller.rb
```diff
@@ -239,14 +239,51 @@ def count_items
...
+    if IssuesForTrayQuery.call(barcode: @tray.barcode).where(issue_type: "incorrect_count").count >= 1
+      true
+    else
+      false
+    end
+  end
+
+  def resolve_count_issues
```
jgondron: Move this to a service class

db/migrate/20160617192442_create_tray_issues.rb
```diff
@@ -0,0 +1,15 @@
+class CreateTrayIssues < ActiveRecord::Migration
```
jgondron: Had you considered just using the existing issues table for this? There is already an issue_type column which could be used to differentiate item issues from tray issues. It's a lot of the same data, just curious if you ruled this out for some reason.

rfox2: I'd much rather handle this in a separate table for a number of reasons. Yerlan had added the issue_type migration in before I had a chance to think about this. One reason to do it this way is to future proof the handling of issues involving separate models. Another is to avoid having to build in complicated logic to build reports or resolve issues. They each have their own workflow, and that's the way I'd like to keep it.

app/controllers/trays_controller.rb
rfox2: Why?

app/controllers/trays_controller.rb
rfox2: Going to the trays with issue report, and clicking on "Resolve" takes the user to the count screen so that they can try again. This is intentional. It follows the workflow as Miranda and Aaron outlined it.

app/controllers/trays_controller.rb
jgondron: It just seems more reasonable to have a query object be responsible for writing and executing this query. For the same reason we pull queries out of Models into query objects, we don't want to bloat a controller with a bunch of query code.

app/controllers/trays_controller.rb
jgondron: This should be handled in a service class object.

db/migrate/20160617192442_create_tray_issues.rb
jgondron: Makes sense

app/controllers/trays_controller.rb
jgondron: Ah I see, I thought resolving these was similar to resolving items with issues.

jgondron:
👍

jonhartzler:
👍

PR #160, opened 2016-09-30
AIMS-320 Add functions to preserve batch search filters

rfox2:
The need was to persist the search filters between page refreshes when the user is creating a batch. I added two callbacks to save, and subsequently load state for custom filter params. The save callback puts the custom filter params into state which is saved locally for the browser. The load callback restores the form controls to the saved state. State is saved / loaded when the form controls are changed. I also added a function that injects a (much needed, IMHO) state load function that can be called as needed.

### DIFF
--- a/app/views/batches/index.html.haml
+++ b/app/views/batches/index.html.haml
@@ -145,21 +145,9 @@
         // set the form elements according to the value of the saved
         // custom params
         "stateLoadParams": function (settings, data) {
-          if (data.columns[1].search.search == "yes") {
-           $('#rapid-check-box').prop("checked", true);
-          } else {
-           $('#rapid-check-box').prop("checked", false);
-          }
-          if (data.columns[2].search.search) {
-            $('#source-select').val(data.columns[2].search.search);
-          } else {
-            $('#source-select').val("aleph");
-          }
-          if (data.columns[3].search.search) {
-            $('#req-type-select').val(data.columns[3].search.search);
-          } else {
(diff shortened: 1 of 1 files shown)


app/views/batches/index.html.haml
```diff
@@ -115,6 +135,32 @@
...
+          if (data.columns[2].search.search) {
+            $('#source-select').val(data.columns[2].search.search);
+          } else {
+            $('#source-select').val("aleph");
+          }
+          if (data.columns[3].search.search) {
+            $('#req-type-select').val(data.columns[3].search.search);
+          } else {
+            $('#req-type-select').val("loan");
+          }
```
hbeachey: This function would be cleaner with something like this instead of the if/else blocks.
```
saveVal = data.search ? data.search : "default"
$('key').val(saveVal)
```

app/views/batches/index.html.haml
```diff
@@ -115,6 +135,32 @@
...
+          data.columns[3].search.search = $('#req-type-select').val();
+        },
+        // set the form elements according to the value of the saved
+        // custom params
+        "stateLoadParams": function (settings, data) {
+          if (data.columns[1].search.search == "yes") {
+           $('#rapid-check-box').prop("checked", true);
+          } else {
+           $('#rapid-check-box').prop("checked", false);
+          }
```
rfox2: Cool. I'll make both of the changes you suggest. All for simplifying code.

PR #162, opened 2016-10-06
Aims 151

hbeachey:
The ticket was a request to search isbn numbers as so: 1234-5678 could be found with 12345678, 3456, 34-56, etc.

To do this efficiently I've changed how items are indexed. This allows us to search for any substring in text searches and affects most search criteria. 
see https://docs.google.com/a/nd.edu/document/d/1Nv-DPRIcrqsJnjT4ifvzmFQA-UOItsJqyj6ZC0Pu1_0/edit?usp=sharing for details


rfox2:
We should definitely get the Travis thing figured out by the end of day tomorrow.

rfox2:
@hbeachey Could you take a look at this new PR with my changes and the previous comments before we merge?

jgondron:
We're losing too much coverage with these spec changes. I think you're either going to have to figure out a way to fix the feature spec, or you're going to have to write unit tests to recoup that coverage.

The search_items_spec has also lost a lot of coverage

jgondron:
Looking at this more closely https://coveralls.io/builds/8240944, I would at least prefer you recoup the missing unit tests on search_items.

rfox2:
So here's the quandary with this and similar problems. The search_items class is filled with private methods. All of the public methods are tested, I think. The more philosophical question is whether or not private methods should be tested. The argument goes both ways. I lean on the side of the luminaries who argue against testing private methods because your unit tests would have more "knowledge" about the class than code that uses the class API. Do the private methods need to be executed for it to be a relevant unit test? Not to get too philosophical and complicate the issue. If this isn't a good set of unit tests, why not? That's the question.

hbeachey:
In general my stance on unit tests is:
- Unit tests should only test _one_ class
- Don't mock/stub methods on the class you're testing, in support of the above point: always mock/stub classes/methods it uses
- Unit tests are _not_ black box tests (integ tests are), they know about the code (they mock/stub pieces)
- All classes should be tested, top to bottom

Now, whether or not these principles are always possible to apply is different than the ideals behind them. For instance, I don't know that it's possible to stub the Item.search do...end block, which is used in every search call. In general this looks good to me though!

Also, I don't know much about haml or if we want to change the hound notifications for those comments.

PR #164, opened 2017-02-20
Items with Issues that won't go away

jgondron:
Its hard to see what's happening to these items in order to identify a pattern since the reports don't show when the issues were resolved (without going into psql and manually querying). Made a few changes to make this easier in the future:
- Added the ResolvedIssue action type to the report for an item.
- Logging of ResolvedIssue action for items is inconsistent with the way other logs are produced. All others expect the associated item to be logged with the action. Changed to require an item when creating a log of this type. Unfortunately, this won't retroactively change logs that are there, so I had to also modify the query to also check if the issue had the item barcode.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -56,7 +56,8 @@ class ItemsController < ApplicationController
 
   def resolve
     issue = Issue.find(params[:issue_id])
-    ResolveIssue.call(user: current_user, issue: issue)
+    item = Item.where(barcode: issue.barcode).take!
+    ResolveIssue.call(item: item, user: current_user, issue: issue)
 
     redirect_to issues_path
   end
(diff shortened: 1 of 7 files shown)


app/controllers/items_controller.rb
```diff
@@ -56,7 +56,8 @@ def issues
 
   def resolve
     issue = Issue.find(params[:issue_id])
-    ResolveIssue.call(user: current_user, issue: issue)
+    item = Item.where(barcode: issue.barcode).take!
```
hbeachey: Do we need to confirm that this item exists before calling `take!`?

jgondron: If this is called with a barcode that does not relate to an item, then something bad happened that the user probably can't correct, so we'll want to see an exception.

PR #170, opened 2017-05-15
Add comments and dispositions to requests

afstanton:
Deaccessioning requests need to store comments and dispositions. This branch fixes that oversight.


afstanton:
Yes, this makes sense.

afstanton:
@rfox2 - yes, that's already in the master branch, in https://github.com/ndlib/annex-ims/blob/master/db/migrate/20170416182126_add_starter_dispositions.rb - so it should already have a couple.

PR #173, opened 2017-05-30
AIMS-409 Call deaccessioning API

afstanton:
Making this call the API and doing it right involves quite a few ancilliary steps along the way.

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -21,7 +21,7 @@ class Item < ActiveRecord::Base
     "error",
   ]
 
-  enum status: { stocked: 0, unstocked: 1, shipped: 2 }
+  enum status: { stocked: 0, unstocked: 1, shipped: 2, deaccessioned: 9 }
 
   validates :thickness, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, presence: true, on: :update
   validates :barcode, uniqueness: true, presence: true
--- a/app/services/deaccession_item.rb
+++ b/app/services/deaccession_item.rb
@@ -19,7 +19,8 @@ class DeaccessionItem
     ApiDeaccessionItemJob.perform_later(item: item)
 
     if item.save!
-      ActivityLogger.deaccession_item(item: item, user: user, disposition: item.disposition)
+      comment = item.try(:requests).order(created_at: :desc).first.try(:comment)
+      ActivityLogger.deaccession_item(item: item, user: user, disposition: item.disposition, comment: { comment: comment } )
       result = item
     else
       result = false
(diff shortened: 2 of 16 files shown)


app/models/item.rb
```diff
@@ -21,7 +21,7 @@ class Item < ActiveRecord::Base
     "error",
   ]
 
-  enum status: { stocked: 0, unstocked: 1, shipped: 2 }
+  enum status: { stocked: 0, unstocked: 1, shipped: 2, deaccessioned: 9 }
```
jgondron: Just curious, why 9?

rfox2: Ditto

afstanton: https://jira.library.nd.edu/browse/AIMS-408 - I'm not sure why, but it was requested to be 9 in the story, so that's what I did.

afstanton:
I think I understand - you want the text of the disposition associated with the item logged (so you know what the value was at the time of deaccessioning, as it could possibly change over time).

afstanton:
Logging the disposition can't be complete until #170 is merged in - I'm working on getting the basic piece in place here.

app/services/deaccession_item.rb
```diff
@@ -0,0 +1,40 @@
...
+  def deaccession!
+    validate_input!
+
+    item.deaccessioned!
+    UpdateIngestDate.call(item)
+    SyncItemMetadata.call(item: item, user_id: nil, background: true)
+    ApiDeaccessionItemJob.perform_later(item: item)
+
+    if item.save!
+      ActivityLogger.deaccession_item(item: item, user: user, disposition: Disposition.first)
```
afstanton: Disposition.first will be replaced with item.disposition once #170 is merged in.

spec/services/deaccession_item_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+RSpec.describe DeaccessionItem do
+  let(:item) { FactoryGirl.create(:item, tray: tray, thickness: 1) }
+  let(:tray) { FactoryGirl.create(:tray) }
+  let(:shelf) { FactoryGirl.create(:shelf) }
+  let(:user) { FactoryGirl.create(:user) }
+  let(:disposition) { FactoryGirl.create(:disposition) }
+  subject { described_class.call(item, user) }
+
+  # This bit is bogus and will be removed after PR #170 gets merged in
```
afstanton: This ensures the existence of at least one Disposition, which is needed for the Disposition.first hack mentioned above. Once #170 is merged in, I'll add disposition to the item factory and this can be removed.

PR #174, opened 2017-06-01
Aims 408 add deaccessioned status

afstanton:
Code seems to work, but I could use some help making the test work.


afstanton:
I agree that the tests need to be fixed, but I could use some help with that, please - I've tried to follow established patterns in testing search results with Solr, but nothing seems to work properly.

afstanton:
I agree that the tests need to be fixed, but I could use some help with that, please - I've tried to follow established patterns in testing search results with Solr, but nothing seems to work properly.

afstanton:
I'm still stuck on this - help would be appreciated.

rfox2:
By the way, my approval alone is OK for this one - there's no way Justin will have time to do code reviews for the next couple of weeks.

rfox2:
Aaron: can you resolve the conflicts on this? I'll look at it ASAP after that. Thanks!

db/migrate/20170531235744_reindex_items.rb
```diff
@@ -0,0 +1,6 @@
+class ReindexItems < ActiveRecord::Migration
```
jgondron: This was already communicated in an email, but just noting it here for anyone else who looks at this PR. This is going to be moved out of a migration and into a rake task that we can perform post deploy.

afstanton:
Migration has been removed. No new rake task seems to be needed, I believe that running `bundle exec rake sunspot:reindex` should handle the required reindex.

PR #200, opened 2018-02-04
Aims 435

afstanton:


afstanton:
@rfox2 - I believe you have it right. The intent is to have the user first enter the old barcode for the item, bring up the data for that, then enter the new barcode and bring up the corresponding new data. After that, the user confirms the merge. It looks to me like this PR is for the middle step of displaying the new barcode data.

rfox2:
Thanks @afstanton - but is the idea in your code that a brand new item record will get created or is the intent just to pull in the metadata for comparison purposes?

PR #209, opened 2018-03-14
Multi pr branch

afstanton:


app/controllers/update_controller.rb
```diff
@@ -29,4 +29,50 @@ def show_old
...
+    end
+
+    if @old_item.blank?
+      flash[:error] = "No item was found with barcode #{params[:old_barcode]}"
+      redirect_to update_path
+      return
+    end
+
+    begin
+      @new_item = GetItemFromBarcode.call(barcode: params[:new_barcode], user_id: current_user.id)
```
rfox2: @afstanton This is the line I have a concern with. From my reading of this class, when this is invoked, the following code path will take place: get() -> item_can_be_stocked?() -> item() -> find_or_create_item() -> Item.find_or_initialize_by() -> item.save!(). If that's the case, it looks like a new item will be created in the db before the user confirms that this is the correct item record. Am I reading that correctly?

app/controllers/update_controller.rb
```diff
@@ -29,4 +29,50 @@ def show_old
...
+      return
+    end
+
+    redirect_to show_new_update_path(old_id: @old_item.id, new_id: @new_item.id)
+    return
+  end
+
+  def show_new
+    @old_item = Item.find(params[:old_id])
+    @new_item = Item.find(params[:new_id])
```
rfox2: New item that was just saved via Active Record?

app/controllers/update_controller.rb
```diff
@@ -29,4 +29,50 @@ def show_old
...
+  end
+
+  def merge
+    MergeNewMetadataToOldItem.call(old_id: params[:old_id], new_id: params[:new_id], user_id: current_user.id)
```
rfox2: I'll have more questions about this class later

app/services/merge_new_metadata_to_old_item.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def merge
+    old_item = Item.find(@old_id)
+    new_item = Item.find(@new_id)
+    user = User.find(@user_id)
+
+    barcode = new_item.barcode
+
+    new_item.barcode = Digest::SHA256.hexdigest(barcode)
+    new_item.save!
```
rfox2: New item record is retained?

app/services/merge_new_metadata_to_old_item.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    old_item.barcode = barcode
```
rfox2: Old item gets new barcode - that makes sense. Then adopts the properties of the new item in order to capture any other metadata changes?

app/controllers/update_controller.rb
afstanton: Yes. I save it in the db because I didn't see any other easy way to store the data between requests.

app/controllers/update_controller.rb
afstanton: Exactly.

app/services/merge_new_metadata_to_old_item.rb
afstanton: I wasn't sure what to do with the new item. I could destroy it here, if that makes more sense to you. I did it this way because destroying data makes me itch - but we can't have two items with the same barcode.

app/services/merge_new_metadata_to_old_item.rb
afstanton: Yes.

afstanton:
I'm okay with destroying the new item after the merge is complete. As mentioned in the code comment, destroying records, even ones that could just be temporary like this one, makes me itch.

It's up to you, please just let me know.

rfox2:
@afstanton I think I'd like to do the metadata lookup for comparison a little differently so that we don't have to create a temporary item record in the db. I'm considering options right now and I'm going to run those by you this week. Today I'm filled up with meetings. Just wanted to give you a heads up.

app/controllers/update_controller.rb
rfox2: I'm not sure that we need to save the data between requests. I'm considering options for this - more later.

afstanton:
I'd like to not create a temporary item if I don't have to. I agree, it clutters things up. I'm not sure how to handle it, though. Maybe some caching service could be used. Let me know after you are done with meetings and such, please. Thanks.

rfox2:
@afstanton I added several comments and code examples that give the idea of the direction I'd like to go in - I'm not sure that we need to create another class to do the fetch because it's only three lines of code to do the fetch and use the response. I'll let you decide on how you want to approach that.

PR #210, opened 2018-04-23
Aims 446 448 449

afstanton:
442 is also in this batch. Please review.

Please do not merge to master until we can deploy to the staging server.

### DIFF
--- a/app/queries/activity_log_query.rb
+++ b/app/queries/activity_log_query.rb
@@ -53,7 +53,7 @@ module ActivityLogQuery
 
   def for_item(item)
     relation.
-      where("data->'issue'->'barcode' ? :barcode OR data->'item'->'barcode' ? :barcode", barcode: item.barcode)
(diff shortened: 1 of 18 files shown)


app/queries/activity_log_query.rb
```diff
@@ -53,7 +53,7 @@ def relation
 
   def for_item(item)
     relation.
-      where("data->'issue'->'barcode' ? :barcode OR data->'item'->'barcode' ? :barcode", barcode: item.barcode)
+      where("data->'issue'->'barcode' ? :barcode OR data->'item'->'barcode' ? :barcode OR data->'item'->'id' ? :id", barcode: item.barcode, id: item.id.to_s)
```
rfox2: @afstanton Just realized - this code as written won't work. It needs to be:

`where("data->'issue'->>'barcode' = ? OR data->'item'->>'barcode' = ? OR data->'item'->>'id' = ?", item.barcode, item.barcode, item.id.to_s)`

PR #211, opened 2018-04-24
Grouping more dependencies into the application group

rfox2:
This is to resolve some issues we're having deploying from Jenkins


afstanton:
Seems reasonable to me.

PR #212, opened 2018-05-01
Fix for tray labels in item detail view

rfox2:
When the full tray object is not stored in the activity log, the valid tray label was not appearing in the item detail view history. This is a fix for that - added a utility method on the query log class to determine what the label should be no matter what is stored in the activity log


afstanton:
This looks good to me!

PR #228, opened 2019-01-19
AIMS-466 ignore hyphens

afstanton:
It may be necessary to do `bundle exec rake sunspot:reindex` after this change is deployed. I've never been able to test Solr with rspec, but this works locally.

I added this to "bin" in addition to trays and shelves because even though bin is not listed in search, we do index on it, so we might want to search on it later.

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -48,7 +48,7 @@ class Item < ActiveRecord::Base
     string :conditions, multiple: true
     date :initial_ingest
     date :last_ingest
-    text :tray_barcode do
+    string :tray_barcode do
       !tray.blank? ? tray.barcode : nil
     end
     text :shelf_barcode do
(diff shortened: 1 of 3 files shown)


app/models/item.rb
```diff
@@ -49,13 +49,13 @@ class Item < ActiveRecord::Base
...
-      !tray.blank? ? tray.barcode : nil
+      !tray.blank? ? tray.barcode.sub(/-/,'') : nil
     end
     text :shelf_barcode do
-      !shelf.blank? ? shelf.barcode : nil
+      !shelf.blank? ? shelf.barcode.sub(/-/,'') : nil
     end
     string :bin_barcode do
-      !bin.blank? ? bin.barcode : nil
+      !bin.blank? ? bin.barcode(/-/,'') : nil
```
rfox2: should this be bin.barcode.sub(/-/,'') ?

afstanton: Yup, I think so.

PR #233, opened 2019-06-08
AIMS-292 Open search results in new tab

afstanton:


afstanton:
No tests added as we have no tests for search results.

PR #234, opened 2019-06-08
AIMS-480 Reposition search field, add confirmation to user creation

afstanton:


afstanton:
I might suggest moving the user creation form to below the search results, too.

PR #241, opened 2019-10-25
Fix the blank error bar in quality check pages

afstanton:
This should resolve the error with an empty error bar showing up on the quality control pages. I also noticed errors persisting past clicking 'done', so I took care of that, too.


rfox2:
@afstanton Do you have the ability to troubleshoot the Travis build? If so, let's try to get that fixed as well.

PR #243, opened 2019-11-15
Rails and ruby upgrade

mlkirkpatrick:
TODO:
- ~Fix last few broken unit tests.~
- ~Upgrade to Rails 5.2.~
- ~Fix hound .rubocop.yml errors.~
- ~Fix Travis (currently failing to install capybara-webkit).~

~@rfox2 Thought you might like to see where this is at. In my testing, the app seemed to be working as expected for everything I tried.~

This updates Ruby to the latest stable release, 2.6.5, and Rails to 5.2.4. It also updates a ton (almost all) of the gems, fixes issues with the Travis build, and issues with Hound/rubocop. There are plenty of linting violations that still should be cleaned up, but at least having it running is a good first step.


afstanton:
It might be easier/better to use different drivers for Capybara than webkit. That's pretty badly out of date, and I'm not sure there are plans to update it.

afstanton:
I'm also curious, why not upgrade to Rails 6?

mlkirkpatrick:
@afstanton That's certainly a possibility. I think main goal was to get it up to a supported version of rails, and with 5.2 we've achieved that. Although it might be worth doing the upgrade to 6.0 while we're at it.

afstanton:
It would buy more time before being forced to upgrade again, I think.

mlkirkpatrick:
The latest stable release of Rspec does not support Rails v6 yet. In light of that, I will not be upgrading to Rails 6 as part of this PR. We can do that later when it is properly supported.

app/services/cancel_batch.rb
```diff
@@ -18,9 +18,7 @@ def cancel!
       request.save!
     end
 
-    batch.matches.each do |match|
-      match.destroy!
-    end
+    batch.matches.each(&:destroy!)
```
rfox2: This syntax seems a little esoteric to me and harder to read than the other - is there a syntactical reason for making this change?

app/services/create_item.rb
```diff
@@ -25,7 +25,10 @@ def create!
       return "errors.barcode_not_found"
     # When the item barcode wasn't found in the database and was set aside by the user
     elsif !set_aside_flag.nil?
-      issue = Issue.find_by(barcode: barcode)
+      issue = Issue.where(barcode: barcode).first_or_create do |new_issue|
```
rfox2: My reading of this is that if an existing issue isn't found, one will be created with this issue type, and then saved again if it was created. Should this be a conditional that creates an issue or saves it but not both?

app/services/is_valid_item.rb
```diff
@@ -2,7 +2,7 @@ module IsValidItem
   def self.call(barcode)
     # This is a terrible way to do this.
     parameter_file = File.join(Rails.root, "config", "barcode_patterns.yml")
-    settings = YAML.load(File.open(parameter_file))
+    settings = YAML.safe_load(File.open(parameter_file), aliases: true)
```
rfox2: What does safe_load do that load doesn't?

config/application.rb
```diff
@@ -24,9 +27,6 @@ class Application < Rails::Application
     # config.i18n.load_path += Dir[Rails.root.join("my", "locales", "*.{rb,yml}").to_s]
     # config.i18n.default_locale = :de
 
-    # Do not swallow errors in after_commit/after_rollback callbacks.
-    config.active_record.raise_in_transactional_callbacks = true
```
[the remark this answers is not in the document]
mlkirkpatrick: It's deprecated and has no effect in Rails 5+.
https://stackoverflow.com/a/37465836/1599426

app/services/shelve_tray.rb
```diff
@@ -23,14 +23,13 @@ def shelve!
...
-  private
+    private
 
-    def validate_input!
-      if IsObjectTray.call(tray)
-        true
-      else
-        raise "object is not a tray"
-      end
+  def validate_input!
```
rfox2: Shouldn't all of this be indented one level?

app/services/ship_item.rb
```diff
@@ -23,17 +23,16 @@ def ship!
...
-  private
+    private
 
-    def validate_input!
-      if IsObjectItem.call(item)
-        true
-      else
-        raise "object is not an item"
-      end
+  def validate_input!
```
rfox2: Indentation?

app/services/unshelve_tray.rb
```diff
@@ -23,14 +23,13 @@ def unshelve!
     end
   end
 
-  private
+    private
```
rfox2: Indentation?

app/services/unstock_item.rb
```diff
@@ -31,14 +31,13 @@ def unstock!
...
-  private
+    private
 
-    def validate_input!
-      if IsObjectItem.call(item)
-        true
-      else
-        raise "object is not an item"
-      end
+  def validate_input!
```
rfox2: Indentation?

config/environments/production.rb
```diff
@@ -14,14 +14,13 @@
...
-  # config.action_dispatch.rack_cache = true
+  # Ensures that a master key has been made available in either ENV["RAILS_MASTER_KEY"]
+  # or in config/master.key. This key is used to decrypt credentials (and other encrypted files).
+  # config.require_master_key = true
 
-  # Disable Rails"s static asset server (Apache or NGINX will already do this).
-  config.serve_static_files = false
+  # Disable serving static files from the `/public` folder by default since
+  # Apache or NGINX already handles this.
+  config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
```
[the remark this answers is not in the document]
mlkirkpatrick: I think I was just following the `app:update` task which happened to use an environment variable along with the rename. We are not actually setting this so I do not believe it's relevant.

app/services/update_ingest_date.rb
```diff
@@ -22,14 +22,13 @@ def update_ingest!
...
-  private
+    private
 
-    def validate_input!
-      if IsObjectItem.call(item)
-        true
-      else
-        raise "object is not an item"
-      end
+  def validate_input!
```
rfox2: This is the last one I'll flag - these look like they need to be indented one level

app/workers/retry_worker.rb
```diff
@@ -7,7 +7,7 @@ def self.from_queue(queue_name, options = {})
     base_options = {
       handler: Sneakers::Handlers::Maxretry,
       arguments: {
-        :"x-dead-letter-exchange" => "#{queue_name}-retry",
+        "x-dead-letter-exchange": "#{queue_name}-retry",
```
rfox2: The syntax translation here isn't clear - is it supposed to return a symbol?

bin/bundle
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env ruby
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
-load Gem.bin_path('bundler', 'bundle')
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("Gemfile", __dir__)
```
rfox2: Just double checking to make sure that this is doing the same thing

config/initializers/sneakers.rb
```diff
@@ -1,8 +1,8 @@
 require "sneakers/handlers/maxretry"
 
 Sneakers.configure(
-  amqp: Rails.application.secrets.sneakers["amqp"],
-  vhost: Rails.application.secrets.sneakers["vhost"],
+  amqp: Rails.application.secrets.sneakers[:amqp],
```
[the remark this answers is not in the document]
mlkirkpatrick: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#application-secrets-now-loaded-with-all-keys-as-symbols

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,8 @@
+# Be sure to restart your server when you modify this file.
```
rfox2: Why is this needed?

config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,25 @@
+# Be sure to restart your server when you modify this file.
```
rfox2: Is this another pro forma required file - do we need it?

config/puma.rb
```diff
@@ -0,0 +1,47 @@
+# Puma can serve each request in a thread from an internal thread pool.
```
rfox2: Is this a switch to Puma for the development environment? Do SSL certs work better with Puma? I know it's a fast container for Rails - just wondering.

db/migrate/20141207174801_create_shelves.rb
```diff
@@ -1,4 +1,4 @@
-class CreateShelves < ActiveRecord::Migration
+class CreateShelves < ActiveRecord::Migration[4.2]
```
rfox2: Is this a version designation that's required because of the way the migrations had been written?

app/controllers/application_controller.rb
```diff
@@ -17,12 +18,12 @@ def check_authentication
     end
     unless user_admin? || user_worker?
       redirect_to_unauthorized
-      return
+      nil
```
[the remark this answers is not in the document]
mlkirkpatrick: I think you're right. Leftover from autocorrect that converted pointless returns into nil instead of removing them altogether.

app/controllers/deaccessioning_controller.rb
```diff
@@ -5,35 +5,36 @@ def index
...
     @results = @res.results
     @total = @res.total
-    @params = params  # Because we need to fill in the form with previous values.
+    @params = params # Because we need to fill in the form with previous values.
     @criteria = params[:criteria]
     @criteria_type = params[:criteria_type]
   end
 
+  def params_whitelist
+    # TODO: :items does not work. In Rails 5.1+, we can do :items => {} to permit a hash as a param.
```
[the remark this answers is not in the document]
mlkirkpatrick: Thanks for the reminder. :)

db/seeds.rb
```diff
@@ -16,7 +16,7 @@ def call_number
   "#{rand_letter}#{rand_letter}#{Faker::Number.number(4)}.#{rand_letter}#{Faker::Number.number(2)} #{(1900..2014).to_a.sample}"
 end
 
-100.times do |i|
+100.times do |_i|
```
rfox2: I've never seen this syntax before - what does it mean?

lib/tasks/annex.rake
```diff
@@ -9,15 +9,15 @@ namespace :annex do
...
   desc "Destroys items created by inaccurate barcode scans or items that were not meant for annex."
   task destroy_invalid_items: :environment do
-    require 'csv'
+    require "csv"
     require Rails.root.join("lib", "tasks", "services", "destroy_invalid_items.rb").to_s
 
     logger = Logger.new(STDOUT)
 
-    system_user = OpenStruct.new({ user_id: nil, username: "system" })
+    system_user = OpenStruct.new(user_id: nil, username: "system")
```
rfox2: I'm only asking this because the syntax for OpenStruct used to **require** curly braces on the constructor - are we sure this is doing the same thing?

spec/controllers/deaccessioning_controller_spec.rb
```diff
@@ -22,34 +22,30 @@
 
   describe "POST req" do
     it "associates an unstocked item with a special bin" do
-      post :req, items: {"#{unstock.id}" => "items[#{unstock.id}]"},
-	   disposition_id: disposition.id,
-           comment: comment
+      post :req, params: { items: { unstock.id.to_s => "items[#{unstock.id}]" }, disposition_id: disposition.id, comment: comment }
```
rfox2: This seems like a really long line - not that we need to stick to 80 characters or less but the other format seemed more readable

app/services/item_restock_to_tray.rb
```diff
@@ -21,24 +21,23 @@ def process!
     tray = GetTrayFromBarcode.call(barcode)
 
     if item.tray.blank? # couldn't figure out how to make link_to work here with doing an include
-        results[:error] = "This item has no tray to stock to."
-        results[:path] = h.show_item_path(:id => @item_id)
+      results[:error] = "This item has no tray to stock to."
+      results[:path] = h.show_item_path(id: @item_id)
```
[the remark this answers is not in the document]
mlkirkpatrick: Well it's using `@item_id` up above to find the item, so I don't think it would really make a difference.

spec/controllers/bins_controller_spec.rb
```diff
@@ -62,20 +64,22 @@
...
+        FactoryBot.create(:match, item: item, bin: bin, request: request2)
         subject
         expect(flash[:warning]).to be_present
       end
 
       it "redirects back to show as worker" do
-        expect(subject).to redirect_to show_bin_path(id: bin.id)
+        subject
+        expected_path = @request.protocol + @request.host + show_bin_path(id: bin.id)
+        expect(response.body).to eq("<html><body>You are being <a href=\"#{expected_path}\">redirected</a>.</body></html>")
```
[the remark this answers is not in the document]
mlkirkpatrick: Probably just me not understanding why the `redirect_to` matcher wasn't working in this case and trying to find something else that worked. :) Admittedly, testing the body is not the best. I can check the status code and the location header instead.

app/jobs/application_job.rb
```diff
@@ -0,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
```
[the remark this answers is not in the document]
mlkirkpatrick: https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#activejob-now-inherits-from-applicationjob-by-default

app/models/application_record.rb
```diff
@@ -0,0 +1,3 @@
+class ApplicationRecord < ActiveRecord::Base
```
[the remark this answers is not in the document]
mlkirkpatrick: Yes. Or at least they prescribe that it should have it even if maybe you could get around it.

app/queries/match_query.rb
```diff
@@ -14,7 +14,7 @@ def retrieve_set(match)
       order("items.title").
       order("items.chron").
       map.
-      with_index { |m, i| set[m.item.id] = "#{i + 1}".to_i.ordinalize }
+      with_index { |m, i| set[m.item.id] = (i + 1).to_s.to_i.ordinalize }
```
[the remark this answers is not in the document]
mlkirkpatrick: Good question. It's not doing anything different, but the new syntax makes the unnecessary transformation much clearer.

app/services/activity_logger.rb
```diff
@@ -199,7 +195,7 @@ def log!
       data: data,
       username: username,
       user_id: user_id,
-      action_timestamp: Time.now,
+      action_timestamp: Time.zone.now,
```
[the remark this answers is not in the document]
mlkirkpatrick: Good catch; I didn't notice this. The linter doesn't like Time.now because it's ambiguous and likely to produce undesired results. It uses the system time, which means if you migrate to a server with a different time zone you won't have consistency. `Time.zone.now` will take into account the Rails environment (e.g. `config.time_zone`), so you can achieve consistency that way.

Given this has historically relied on `Time.now`, I can just revert these changes to prevent an impact to logging. If I knew what the servers are running on, potentially we could resolve this by either using utc or by ensuring the Rails time zone is set to match what has been used up until now.

spec/controllers/deaccessioning_controller_spec.rb
jeremyf: One option would be to extract the `params` value into a variable on the above line. Oddly, is it important that the `unstock.id` be a string?

app/services/create_item.rb
mlkirkpatrick: You're probably right. I think I was trying to fix a situation where an issue does not exist, but updating it after create is unnecessary.

spec/controllers/deaccessioning_controller_spec.rb
mlkirkpatrick: Agreed

spec/workers/retry_worker_spec.rb
```diff
@@ -29,7 +29,7 @@ class DefaultsWorker < RetryWorker
     end
 
     it "sets the correct queue options" do
-      expect(queue_options[:arguments]).to eq(:"x-dead-letter-exchange" => "testdefaults-retry")
+      expect(queue_options[:queue_options][:arguments]).to eq("x-dead-letter-exchange": "testdefaults-retry")
```
jeremyf: This looks a bit surprising, shifting from `queue_options[:arguments]` to `queue_options[:queue_options][:arguments]`.

spec/workers/item_metadata_worker_spec.rb
```diff
@@ -23,7 +23,7 @@
     end
 
     it "sets the correct queue options" do
-      expect(queue_options[:arguments]).to eq(:"x-dead-letter-exchange" => "annex_item_metadata-retry")
+      expect(queue_options[:queue_options][:arguments]).to eq("x-dead-letter-exchange": "annex_item_metadata-retry")
```
jeremyf: This looks a bit surprising, shifting from queue_options[:arguments] to queue_options[:queue_options][:arguments].

spec/workers/api_worker_spec.rb
```diff
@@ -23,7 +23,7 @@
     end
 
     it "sets the correct queue options" do
-      expect(queue_options[:arguments]).to eq(:"x-dead-letter-exchange" => "annex_api-retry")
+      expect(queue_options[:queue_options][:arguments]).to eq("x-dead-letter-exchange": "annex_api-retry")
```
jeremyf: This looks a bit surprising, shifting from queue_options[:arguments] to queue_options[:queue_options][:arguments].

lib/tasks/annex.rake
mlkirkpatrick: That does not seem to be the case anymore.
https://ruby-doc.org/stdlib-2.5.1/libdoc/ostruct/rdoc/OpenStruct.html

spec/services/withdraw_tray_spec.rb
```diff
@@ -28,7 +28,7 @@
   end
 
   it "registers the call with the transaction log" do
```
jeremyf: As this is a commented out test, perhaps switch from `it` to `xit`, which will mark the spec as pending (instead of the current success with no assertion)

spec/services/associate_tray_with_shelf_barcode_spec.rb
```diff
@@ -34,10 +33,9 @@
     expect(subject).to be(false)
   end
 
-
-# this is not implemented in the class..
+  # this is not implemented in the class..
   it "raises an error if the item is not a tray." do
```
jeremyf: As this is a commented out test, perhaps switch from it to xit, which will mark the spec as pending (instead of the current success with no assertion)

jeremyf: As this is a commented out test, perhaps switch from it to xit, which will mark the spec as pending (instead of the current success with no assertion)

spec/workers/retry_worker_spec.rb
mlkirkpatrick: Indeed, but that's how the object is structured. I don't know why it changed, but `:arguments` only exists as a child of `:queue_options`. The naming redundancy makes it a bit confusing. In this case, queue_options is `queue.opts.to_hash`, so technically the path is `queue.opts[:queue_options][:arguments]`

spec/services/withdraw_tray_spec.rb
mlkirkpatrick: Thanks for the suggestion!

spec/services/unshelve_tray_spec.rb
```diff
@@ -31,10 +31,9 @@
     expect(subject).to be(false)
   end
 
-
-# this is not implemented in the class..
+  # this is not implemented in the class..
   it "raises an error if the item is not a tray." do
```
jeremyf: As this is a commented out test, perhaps switch from it to xit, which will mark the spec as pending (instead of the current success with no assertion)

db/migrate/20141207174801_create_shelves.rb
mlkirkpatrick: There were some API changes to the way migrations work in Rails 5. To ensure old migrations are not broken, there's a compatibility layer that allows them to run if they are versioned correctly. 4.2 is the first release that has versioning, so everything that was generated prior to the rails 5 update should be marked as 4.2.

If a migration does not have a version I believe it will still be treated as if it were 4.2, but it will also produce a deprecation warning so it's best to be explicit. :)

https://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html

config/puma.rb
mlkirkpatrick: Not actually using it. This was just generated by the `rails app:update` task.

Guardfile
```diff
@@ -23,9 +23,9 @@ hesburgh_guard.rails do
   # Watch any custom paths
 end
 
-guard 'sunspot', :environment => 'development' do
```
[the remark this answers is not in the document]
mlkirkpatrick: I don't know, you tell me. :) I'm all for getting rid of stuff we don't need, but I also don't know what guard is or how it ties in.

app/controllers/dispositions_controller.rb
```diff
@@ -18,8 +18,7 @@ def new
     respond_with(@disposition)
   end
 
-  def edit
-  end
+  def edit; end
```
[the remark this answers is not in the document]
mlkirkpatrick: I suspect we do not! Consider it gone.

app/services/cancel_batch.rb
mlkirkpatrick: It's just elegant and the style guide says so.
https://github.com/rubocop-hq/ruby-style-guide#single-action-blocks
https://www.brianstorti.com/understanding-ruby-idiom-map-with-symbol/

The syntax is a little different, but for what it's worth, I would also prefer a single-line solution in javascript:
```
batch.matches.forEach(match => match.destroy())
```
vs
```
for (const match in batch.matches) {
  match.destroy()
}
```

app/services/is_valid_item.rb
mlkirkpatrick: From my understanding: `YAML.load` allows importing an arbitrary data structure, which can potentially instantiate class instances. `YAML.safe_load` restricts this to built-in types, or those explicitly permitted. The file we're importing is on the server side, so there really isn't much of a risk, but there's also no harm in using the safer method and it makes the linter happy.

https://ruby-doc.org/stdlib-2.6.1/libdoc/psych/rdoc/Psych.html#method-c-load
https://ruby-doc.org/stdlib-2.6.1/libdoc/psych/rdoc/Psych.html#method-c-safe_load

bin/bundle
mlkirkpatrick: Bundler works. I don't know how it could be if it wasn't.

config/initializers/content_security_policy.rb
mlkirkpatrick: It seems it's just another file Rails makes to let you know it's a feature that's available. I'd be fine deleting it, but I also wonder if that might go against the "Rails-way" of doing things. There's a ton of configuration options which come commented out, and if we are going to remove files because we are not using those features, it seems like we should be consistent and do the same across the entire project.

db/seeds.rb
mlkirkpatrick: It doesn't do anything special to the code - it's just to make the linter happy.

In this case, `i` is unused. Omitting it might be valid, although might look a bit goofy:
`100.times do ||`
In this scenario, we don't actually need the index, but listing it as a variable serves a visual queue to the developer so they know what is coming in. This could be useful in more complex loops where you might have multiple arguments, but only a few of them are used at a particular time.

The underscore just tells the linter to ignore the unused variable because you don't intend to use it in the block, but saw value in listing it anyway.

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/UnusedBlockArgument

app/workers/retry_worker.rb
mlkirkpatrick: It is still a symbol. The quotes are only necessary because of the dashes.

Guardfile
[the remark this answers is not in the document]
rfox2: Guard is used in development to watch for code changes - when a code change happens it can run scoped unit tests against the code that changed so that you get up to the minute unit test results but they haven't been maintaining the gem for a long time. I think we can extract Guard from the IMS.

config/initializers/content_security_policy.rb
rfox2: No we can keep the stub - no problem

app/services/activity_logger.rb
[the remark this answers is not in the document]
rfox2: OK, no problem. There were several places in your PR where it was changed to Time.zone.now - are you going to revert all of those?

app/services/is_valid_item.rb
rfox2: Neat - I didn't know about that

app/services/shelve_tray.rb
rfox2: There were several places in the code where the indent looked off - did you fix all of those as well?

rfox2:
If there are empty methods in controllers - and we don't need that method for some reason - make sure that the routes file doesn't reference them. It could be that the empty controller methods were there because we were using the implicit "resources" syntax that automatically creates certain routes for CRUD operations.

rfox2:
Also, are we going to fix the Travis build? Please? ;-)

app/services/cancel_batch.rb
jeremyf: The `collection.each(&:destroy)` is a very common syntax for Ruby/Rails, that was introduced in Ruby 2.0.0

jeremyf:
@rfox2 It is possible with empty methods that Rails does some work (e.g. will auto-render the correct HTML page). However, if there is no corresponding `new.html` page then that could be removed.

PR #257, opened 2020-06-08
Renaming whitelist to allowed

jeremyf:
Prior to this commit, our language of Whitelist perpetuated the idea
that "white" is _the_ allowed. This change clarifies the purpose without
relying on discriminatory cultural jargon (e.g. blacklist/whitelist). In
addition, by using a clear term, we make the intention more accessible.

### DIFF
--- a/app/controllers/deaccessioning_controller.rb
+++ b/app/controllers/deaccessioning_controller.rb
@@ -10,7 +10,7 @@ class DeaccessioningController < ApplicationController
     @criteria_type = params[:criteria_type]
   end
 
-  def params_whitelist
+  def params_allowed
     params.permit(items: {})
     params.permit([:disposition_id, :comment])
   end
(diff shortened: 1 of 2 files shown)


rfox2:
@jeremyf I understand that you feel passionately about this. And there are ways for you to support your passions outside of work. But to bring politically motivated commentary into how we write our code is entirely inappropriate. This use "white" or "black" has nothing to do with race in any sense. I am completely opposed to the idea of regulating language in our technical work to support **any** political or religious position.

jeremyf:
Blacklist / whitelist is more opaque than deny / allow language. This is not political, this is about clarity while also slowly correcting language that perpetuates systemic inequities.

ialford:
I would agree with @jeremyf on the opacity of whitelist/blacklist vs allow/deny. 

@rfox2, if it wasn't for the mention of `discriminatory cultural jargon` in the PR request, what objections would there be to the PR? There are certainly concerns with changing parameter names arbitrarily, as it could introduce unforeseen issues. 

We should also consider that this as an open source project and has a Code of Conduct, which version of the code is more in line with our CoC?

rfox2:
"blacklist" and "whitelst" have been used in IT for ages. This PR is politically motivated and done for a personal agenda. I can't approve it because I think it sets a **very bad** standard for us. Jeremy clearly indicated that his motivation for this PR is politically motivate in his support of the current cultural conditions. I have to strongly oppose that and I don't think it has any place at work. If we allowed that, it would be a free-for-all. Whitelist and blacklist have been used all over the place in IT language - it has an established understanding and has nothing to do with discrimination.

rfox2 (changes requested):
Tried to discuss some reservations about this pull request - accepted IT language vs. debatable social implications

mdehmlow:
Hi all, this is clearly an issue that people feel strongly about - me included.  This is probably not the forum where we can have a meaningful discussion about this.  I will schedule a time for us to discuss where we can have a constructive conversation.  I want to re-emphasize that it is important to treat each other with respect - especially when we feel strongly or disagree, otherwise we erode trust, stifle communication, and create an uncomfortable climate.

PR #260, opened 2021-01-11
Add CI via GitHub actions

ialford:
Adding a unit test CI runner via GitHub Actions. A couple of things here:

- The runner had to be downgraded to Ubuntu 18.04, which is one LTS version behind, due to lemurheavy/coveralls-ruby#163. This could be remedied by changing gems to coveralls-ruby-reborn if that is a desired action, but it is beyond the scope of the PR.
- A new "template" database config file was created for the Actions runner, so as to not interfere with any of the active development that may impact either the .travis file or the main database.yml config file.
- We're only running this for Ruby 2.6 currently. It may make sense to add 2.7 as a sanity check for a future minor upgrade.
- A few unit tests are failing. Fixing these is beyond the scope of the PR - we can still merge as long as successful checks aren't required. 

Feel free to ask any questions if they arise.


rfox2 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+
+      # Install gems
+      - name: Install dependencies
+        run: bundle install
+
+      # Put the database and secrets files in place
+      - name: Create secrets.yml and database.yml
+        run: |
+            cp config/secrets.yml.travis config/secrets.yml
+            cp config/database.yml.travis config/database.yml
```
mlkirkpatrick, line 54: Don't you want the `.actions` file instead of `.travis`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,70 @@
...
+        
+      # Create the test database
+      - name: Create test database
+        env:
+          PGPASSWORD: password
+          PGHOST: localhost
+          PGUSER: postgres
+        run: psql -c 'create database travis_ci_test;' -U postgres
```
mlkirkpatrick, line 62: Would also change the name of the database to be more accurate if it doesn't interfere with the ability to run the tests.

mlkirkpatrick (changes requested):

.github/workflows/ci.yml
ialford: Fixed, good catch

.github/workflows/ci.yml
ialford: When I changed the DB name, I had errors in running rake against the DB because it couldn't find the `travis_ci_test` database, so I think it's needed.

mlkirkpatrick: I imagine it should work as long as you update the database name for the test environment in `database.yml.actions`. Maybe it was failing because it was still pointing to the travis file which had the old name?

mlkirkpatrick (approved):

PR #273, opened 2021-10-01
task: bring annex-ims script over from dmprod (das-2873)

msuhovec:


scripts/request.csh
```diff
@@ -0,0 +1,22 @@
...
+#  the Annex
+##################################
+
+if ($1 == '' ) then
+    set prod = 'prod'
+else
+    set prod = $1
+endif
+
+set annexpath = "/api-gateway/annex/$prod/outbound"
```
rdoughty, line 16:
```suggest
if ("$1" == "" ) then
    set stage = 'prod'
else
    set stage = $1
endif

set annexpath = "/api-gateway/annex/${stage}/outbound"

rdoughty (approved):

